PR #5, opened 2017-12-18
Sanity Test Suite: Initial Framework

lpabon:
This is just the beginning of the framework. But I wanted to get acceptance on its architecture before continuing.

Currently, it only tests the Identity Server.


lpabon:
Implementes framework discussed in https://github.com/kubernetes-csi/drivers/issues/10

PR #24, opened 2018-01-23
Support Unix Domain Socket for CSI Test

edisonxiang:
* add uds support fo csi test.
* remove `.` from logs.

### DIFF
--- a/cmd/csi-sanity/sanity_test.go
+++ b/cmd/csi-sanity/sanity_test.go
@@ -45,7 +45,7 @@ func TestSanity(t *testing.T) {
 		return
 	}
 	if len(endpoint) == 0 {
-		t.Fatalf("--%s.endpoint must be provided with an CSI endpoint", prefix)
+		t.Fatalf("--%sendpoint must be provided with an CSI endpoint", prefix)
 	}
 	sanity.Test(t, endpoint)
 }
--- a/utils/grpcutil.go
+++ b/utils/grpcutil.go
@@ -1,0 +1,42 @@
+/*
+Copyright 2017 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package utils
+
+import (
+	"net"
+	"net/url"
+	"time"
+
+	"google.golang.org/grpc"
+)
+
+// Connect address by grpc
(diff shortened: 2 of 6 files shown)


cmd/csi-sanity/sanity_test.go
```diff
@@ -45,7 +45,8 @@ func TestSanity(t *testing.T) {
 		return
 	}
 	if len(endpoint) == 0 {
-		t.Fatalf("--%s.endpoint must be provided with an CSI endpoint", prefix)
+		t.Fatalf("--%sendpoint must be provided with an CSI endpoint", prefix)
```
lpabon: It needs to be `%s.endpoint` because all help has a period. For example there is `ginkgo.XXX` and `test.XXX`

cmd/csi-sanity/sanity_test.go
```diff
@@ -45,7 +45,8 @@ func TestSanity(t *testing.T) {
...
 	}
+	fmt.Printf("Endpoint = %s\n", endpoint)
```
lpabon: Do not use fmt. This is more like a unit test. Use this instead if you want: http://onsi.github.io/ginkgo/#logging-output .


I feel it is not really necessary to print out the endpoint, but I wouldn't be against only printing it on verbose output.

utils/grpcutil.go
```diff
@@ -0,0 +1,49 @@
...
+	"net"
+	"regexp"
+	"strings"
+	"time"
+
+	"google.golang.org/grpc"
+)
+
+// getProtoandAdd return protocal and address
+func getProtoandAdd(target string) (string, string) {
```
lpabon: I don't understand the purpose of this function. As far as I know, this all works with unix domain sockets (That's how I test it). Could you explain why this is necessary?

lpabon:
Actually, now that I think about it, I haven't tested with unix domain socket. I will do that first.

lpabon:
I am not sure if you noticed, but I removed my commend I had made on the `%sendpoint` change. I just noticed:

```
$ ./csi-sanity 
--- FAIL: TestSanity (0.00s)
        sanity_test.go:48: --csi..endpoint must be provided with an CSI endpoint
FAIL
```

You are correct. good find.

lpabon:
@edisonxiang Just proved that it didn't work with unix domain sockets as you proposed. Thanks for finding this.

Now I am looking into your code and checking to see how it compares with driver-registrar, external-provisioner, and external-attacher.

lpabon:
I have finally checked out the issue. Thank you very much for the help. I would like to propose a different proposal for Connect:

cmd/csi-sanity/sanity_test.go
edisonxiang: Thanks. I will remote Line 50

edisonxiang:
@lpabon Thanks.

edisonxiang:
CSI_ENDPOINTS="tcp://127.0.0.1:9998"
CSI_ENDPOINTS="/tmp/csi.sock"
CSI_ENDPOINTS="unix:///tmp/csi.sock"

lpabon:
🎉

PR #31, opened 2018-01-24
sanity: CreateVolume

lpabon:
This change has seems to have found an error in the mock driver,
so until we find out the issue, the tests will be skipped in
`hack/e2e.sh` with `--ginkgo.skip=MOCKERRORS`.

Also introduced more reliable connection to the CSI driver
by only creating a single connection once during the test.

Signed-off-by: Luis Pabón <luis@portworx.com>


lpabon:
Closes #18

PR #63, opened 2018-03-26
Changed Mkdir to create full directory. Also creates Staging Path

davidz627:
Mkdir will fail when required parent path does not exist. MkdirAll creates the full directory structure required by Target and Staging path.

### DIFF
--- a/pkg/sanity/sanity.go
+++ b/pkg/sanity/sanity.go
@@ -63,6 +63,10 @@ var _ = BeforeSuite(func() {
 	By("creating mount and staging directories")
 	err = createMountTargetLocation(config.TargetPath)
 	Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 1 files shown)


pkg/sanity/sanity.go
```diff
@@ -63,6 +63,8 @@ var _ = BeforeSuite(func() {
 	By("creating mount and staging directories")
 	err = createMountTargetLocation(config.TargetPath)
 	Expect(err).NotTo(HaveOccurred())
+	err = createMountTargetLocation(config.StagingPath)
```
lpabon: What if staging path is not needed and it is not set. We should check here and avoid this section, right?

davidz627: resolved

davidz627:
comments addressed

PR #64, opened 2018-03-27
Testing with 'RequiredBytes' should use it as a minimum value according to spec

davidz627:
The Spec on required bytes:
  // Volume must be at least this big. This field is OPTIONAL.
  // A value of 0 is equal to an unspecified field value.
  int64 required_bytes = 1;

Therefore tests should accept any CreateVolume return size greater than or equal to the required bytes size.


davidz627:
@lpabon @sbezverk PTAL

PR #66, opened 2018-04-16
sanity: Add tests for NodeStageVolume/NodeUnstageVolume

### LINKED ISSUE #46: Tests for NodeStageVolume/NodeUnstageVolume
Tests do not currently exist for NodeStageVolume/NodeUnstageVolume. They should be added to `pkg/sanity/node.go`

This TODO was introduced by #43

darkowlzz:
Fixes #46

### DIFF
--- a/mock/service/node.go
+++ b/mock/service/node.go
@@ -110,6 +110,18 @@ func (s *service) NodePublishVolume(
 			"publish volume info 'device' key required")
 	}
 
+	if len(req.GetVolumeId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume ID cannot be empty")
+	}
+
+	if len(req.GetTargetPath()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Target Path cannot be empty")
+	}
+
+	if req.GetVolumeCapability() == nil {
+		return nil, status.Error(codes.InvalidArgument, "Volume Capability cannot be empty")
+	}
+
 	s.volsRWL.Lock()
 	defer s.volsRWL.Unlock()
 
@@ -135,7 +147,11 @@ func (s *service) NodePublishVolume(
 	}
 
 	// Publish the volume.
-	v.Attributes[nodeMntPathKey] = device
+	if req.GetStagingTargetPath() != "" {
+		v.Attributes[nodeMntPathKey] = req.GetStagingTargetPath()
+	} else {
+		v.Attributes[nodeMntPathKey] = device
+	}
 	s.vols[i] = v
 
 	return &csi.NodePublishVolumeResponse{}, nil
--- a/mock/service/node.go
+++ b/mock/service/node.go
@@ -16,7 +16,47 @@ func (s *service) NodeStageVolume(
 	req *csi.NodeStageVolumeRequest) (
 	*csi.NodeStageVolumeResponse, error) {
 
-	return nil, status.Error(codes.Unimplemented, "")
+	device, ok := req.PublishInfo["device"]
+	if !ok {
+		return nil, status.Error(
+			codes.InvalidArgument,
+			"stage volume info 'device' key required")
+	}
+
+	if len(req.GetVolumeId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Volume ID cannot be empty")
+	}
+
+	if len(req.GetStagingTargetPath()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Staging Target Path cannot be empty")
+	}
+
+	if req.GetVolumeCapability() == nil {
+		return nil, status.Error(codes.InvalidArgument, "Volume Capability cannot be empty")
+	}
+
+	s.volsRWL.Lock()
+	defer s.volsRWL.Unlock()
+
+	i, v := s.findVolNoLock("id", req.VolumeId)
+	if i < 0 {
+		return nil, status.Error(codes.NotFound, req.VolumeId)
+	}
+
+	// nodeStgPathKey is the key in the volume's attributes that is set to a
+	// mock stage path if the volume has been published by the node
+	nodeStgPathKey := path.Join(s.nodeID, req.StagingTargetPath)
+
+	// Check to see if the volume has already been staged.
+	if v.Attributes[nodeStgPathKey] != "" {
+		return &csi.NodeStageVolumeResponse{}, nil
+	}
+
+	// Stage the volume.
+	v.Attributes[nodeStgPathKey] = device
+	s.vols[i] = v
+
+	return &csi.NodeStageVolumeResponse{}, nil
 }
 
 func (s *service) NodeUnstageVolume(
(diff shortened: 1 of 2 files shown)


pkg/sanity/node.go
```diff
@@ -364,3 +364,328 @@ func testFullWorkflowSuccess(s csi.ControllerClient, c csi.NodeClient, controlle
...
+		s = csi.NewControllerClient(conn)
+		c = csi.NewNodeClient(conn)
+		controllerPublishSupported = isControllerCapabilitySupported(
+			s,
+			csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME)
+		nodeStageSupported = isNodeCapabilitySupported(c, csi.NodeServiceCapability_RPC_STAGE_UNSTAGE_VOLUME)
+		if nodeStageSupported {
+			err := createMountTargetLocation(config.StagingPath)
+			Expect(err).NotTo(HaveOccurred())
+		}
```
lpabon: If `STAGE_UNSTAGE_VOLUME` is not supported it should skip this test.

mock/service/node.go
```diff
@@ -16,15 +16,84 @@ func (s *service) NodeStageVolume(
 	req *csi.NodeStageVolumeRequest) (
 	*csi.NodeStageVolumeResponse, error) {
 
-	return nil, status.Error(codes.Unimplemented, "")
+	device, ok := req.PublishInfo["device"]
```
lpabon: To support this, `STAGE_UNSTAGE_VOLUME` needs to be supported in https://github.com/kubernetes-csi/csi-test/blob/master/mock/service/node.go#L119 and the test should skip if not supported.

pkg/sanity/node.go
lpabon: See https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/controller.go#L105 as an example.

mock/service/node.go
```diff
@@ -16,15 +16,84 @@ func (s *service) NodeStageVolume(
...
+	if i < 0 {
+		return nil, status.Error(codes.NotFound, req.VolumeId)
+	}
+
+	// nodeStgPathKey is the key in the volume's attributes that is set to a
+	// mock stage path if the volume has been published by the node
+	nodeStgPathKey := path.Join(s.nodeID, req.StagingTargetPath)
+
+	// Check to see if the volume has already been unstaged.
+	if v.Attributes[nodeStgPathKey] == "" {
```
lpabon: This assumes that the key is there. It could panic. There needs to be a check

mock/service/node.go
```diff
@@ -16,15 +16,84 @@ func (s *service) NodeStageVolume(
...
+	if i < 0 {
+		return nil, status.Error(codes.NotFound, req.VolumeId)
+	}
+
+	// nodeStgPathKey is the key in the volume's attributes that is set to a
+	// mock stage path if the volume has been published by the node
+	nodeStgPathKey := path.Join(s.nodeID, req.StagingTargetPath)
+
+	// Check to see if the volume has already been staged.
+	if v.Attributes[nodeStgPathKey] != "" {
```
lpabon: This assumes the key is in the map and could panic. It needs to get the key by also checking to see if it is there.

mock/service/node.go
```diff
@@ -16,15 +16,84 @@ func (s *service) NodeStageVolume(
...
+		return &csi.NodeStageVolumeResponse{}, nil
```
davidz627: Should check whether the volume staged is compatible. See error "ALREADY_EXISTS" in spec.

mock/service/node.go
```diff
@@ -16,15 +16,84 @@ func (s *service) NodeStageVolume(
...
+	}
+
+	// Unpublish the volume.
+	delete(v.Attributes, nodeStgPathKey)
+	s.vols[i] = v
+
+	return &csi.NodeUnstageVolumeResponse{}, nil
 }
 
 func (s *service) NodePublishVolume(
```
davidz627: Need to modify NodePublishVolume to do the right thing when a StagingTargetPath is provided.

pkg/sanity/node.go
```diff
@@ -364,3 +364,328 @@ func testFullWorkflowSuccess(s csi.ControllerClient, c csi.NodeClient, controlle
...
+	})
+
+	It("should fail when no volume id is provided", func() {
+
+		_, err := c.NodeStageVolume(
+			context.Background(),
+			&csi.NodeStageVolumeRequest{})
```
davidz627: provide the rest of the required fields to isolate the error as just a volume id error

pkg/sanity/node.go
```diff
@@ -364,3 +364,328 @@ func testFullWorkflowSuccess(s csi.ControllerClient, c csi.NodeClient, controlle
...
+		Expect(ok).To(BeTrue())
+		Expect(serverError.Code()).To(Equal(codes.InvalidArgument))
+	})
+
+	It("should fail when no staging target path is provided", func() {
+
+		_, err := c.NodeStageVolume(
+			context.Background(),
+			&csi.NodeStageVolumeRequest{
+				VolumeId: "id",
```
davidz627: provide the rest of the required fields.

pkg/sanity/node.go
```diff
@@ -364,3 +364,328 @@ func testFullWorkflowSuccess(s csi.ControllerClient, c csi.NodeClient, controlle
...
+			})
+		Expect(err).To(HaveOccurred())
+
+		serverError, ok := status.FromError(err)
+		Expect(ok).To(BeTrue())
+		Expect(serverError.Code()).To(Equal(codes.InvalidArgument))
+	})
+
+	It("should return appropriate values (no optional values added)", func() {
```
davidz627: Is this similar enough to `testFullWorkflowSuccess` that we could reuse that - or make some minimal tweaks to that function and reuse it here?

pkg/sanity/node.go
```diff
@@ -364,3 +364,328 @@ func testFullWorkflowSuccess(s csi.ControllerClient, c csi.NodeClient, controlle
...
+				StagingTargetPath: config.StagingPath,
+			})
+		Expect(err).To(HaveOccurred())
+
+		serverError, ok := status.FromError(err)
+		Expect(ok).To(BeTrue())
+		Expect(serverError.Code()).To(Equal(codes.InvalidArgument))
+	})
+
+	It("should return appropriate values (no optional values added)", func() {
```
davidz627: Is this similar enough to testFullWorkflowSuccess that we could reuse that - or make some minimal tweaks to that function and reuse it here?

https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/node.go#L224

pkg/sanity/node.go
```diff
@@ -364,3 +364,328 @@ func testFullWorkflowSuccess(s csi.ControllerClient, c csi.NodeClient, controlle
...
+		nodeStageSupported         bool
+	)
+
+	BeforeEach(func() {
+		s = csi.NewControllerClient(conn)
+		c = csi.NewNodeClient(conn)
+		controllerPublishSupported = isControllerCapabilitySupported(
+			s,
+			csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME)
+		nodeStageSupported = isNodeCapabilitySupported(c, csi.NodeServiceCapability_RPC_STAGE_UNSTAGE_VOLUME)
```
davidz627: skip test if nodeStage not supported

pkg/sanity/node.go
```diff
@@ -364,3 +364,328 @@ func testFullWorkflowSuccess(s csi.ControllerClient, c csi.NodeClient, controlle
...
+			err := createMountTargetLocation(config.StagingPath)
+			Expect(err).NotTo(HaveOccurred())
+		}
+	})
+
+	It("should fail when no volume id is provided", func() {
+
+		_, err := c.NodeUnstageVolume(
+			context.Background(),
+			&csi.NodeUnstageVolumeRequest{})
```
davidz627: ditto: provide rest of the required args

pkg/sanity/node.go
```diff
@@ -364,3 +364,328 @@ func testFullWorkflowSuccess(s csi.ControllerClient, c csi.NodeClient, controlle
...
+		Expect(ok).To(BeTrue())
+		Expect(serverError.Code()).To(Equal(codes.InvalidArgument))
+	})
+
+	It("should fail when no staging target path is provided", func() {
+
+		_, err := c.NodeUnstageVolume(
+			context.Background(),
+			&csi.NodeUnstageVolumeRequest{
+				VolumeId: "id",
```
davidz627: ditto: provide rest of the required args

mock/service/node.go
darkowlzz: I tried adding this, but this would require changes in service.go, in the `Volume` struct. I think it would be better to add separately, capability check for Controller functions as well. I'll add a TODO for now here.

pkg/sanity/node.go
darkowlzz: Yes, it's completely same. I failed to see that :)

mock/service/node.go
```diff
@@ -64,7 +140,7 @@ func (s *service) NodePublishVolume(
 	}
 
 	// Publish the volume.
-	v.Attributes[nodeMntPathKey] = device
+	v.Attributes[nodeMntPathKey] = req.GetStagingTargetPath()
```
darkowlzz: Changed this from device to the staging target path. I think staging path is bind mounted to the target path. Correct me if I'm wrong about that.

lpabon: I'm not sure. I'll ask.

davidz627: this should only be the staging target path if it is defined - if not, it would be device.

darkowlzz: Right. Thanks!

darkowlzz:
Made the changes. Hope it's ready.

mock/service/node.go
lpabon: You are right! That must be new since I started using golang (1.2): https://play.golang.org/p/OMC7Taio5h1

lpabon:
Looks great

PR #74, opened 2018-06-04
sanity: configurable volume size

pohly:
Not all CSI drivers necessarily support 10GB volumes. Some simplistic,
memory-backed test drivers for example might only work with some
smaller size.


lpabon:
Good idea, can you also add it to as a cmd/csi-sanity flag?

pohly:
> Good idea, can you also add it to as a cmd/csi-sanity flag?

Done.

lpabon:
@pohly FYI, you just need now to resolve the conflicts.

pohly:
Conflicts resolved.

Issue #73, opened 2018-06-01
pkg/sanity usage as non-root user


pohly:
When following https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/README.md and running "go test" as normal user (!), tests involving mounting/unmounting fail because those are privileged operations.

Running "sudo go test" isn't a good solution, because then building also runs as root. This is not desirable and can also fail because the environment variables for Go end up being different.

My workaround was to have mount/unmount wrapper scripts that run just the mount and unmount command under pseudo. I can submit a PR with just those scripts (see below), but then one still has to remember to update PATH.

I wonder whether this isn't something that the test framework itself should take care of. Any thoughts?

```
#!/bin/sh

# Get rid of our wrapper script by setting a standard search PATH.
PATH=/usr/bin:/bin:/usr/sbin:/sbin

if [ $(id -u) != 0 ]; then
   exec sudo $(basename $0) "$@"
else
   exec $(basename $0) "$@"
fi
```

lpabon:
If this is run from `go test` then as a user I would suggest building a binary and using that as sudo. csi-test should not know what user it is used as.

pohly:
Luis Pabón <notifications@github.com> writes:
> If this is run from `go test` then as a user I would suggest building
> a binary and using that as sudo.

I was aware of that option and found it that rather inconvenient. "dlv
test" also doesn't work out-of-the-box.

But okay, let's close this.

PR #75, opened 2018-06-05
sanity: mount via sudo

pohly:
"go test" is better called as normal user because then the build
environment and resulting files are for that user instead of root. But
usually mounting volumes will then fail because it is a privileged
operation and/or the necessary tools aren't in the default search path
of a normal user.

This can be avoided by wrapping the tools that might be needed (for
example, by the Kubernetes mount utility code) such that they get
invoked via sudo. If the CSI driver under test does not need them,
nothing changes, but if it does, the test now might succeed.

Fixes: kubernetes-csi/csi-test#73

### DIFF
--- a/pkg/sanity/node.go
+++ b/pkg/sanity/node.go
@@ -18,6 +18,9 @@ package sanity
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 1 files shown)


pohly:
This works for a CSI driver that I am currently working on. If it's not deemed useful enough for csi-test, then I can also put the code into my own _test.go function. But to me it looked like something that others might also find useful.

pohly:
It should also work for the hostpath driver, thus making it possible to remove the e2e-hostpath.sh script and replace it with a normal go test file.

The only caveat then is that the test file must be able to skip unsupported tests, which probably will require some more entries in csi-test Config.

PR #78, opened 2018-06-15
Mock driver: Return probe ready status

darkowlzz:
CSI 0.3.0 introduced probe ready status. This change adds the Ready
value in ProbeResponse.

### DIFF
--- a/pkg/sanity/identity.go
+++ b/pkg/sanity/identity.go
@@ -79,6 +79,11 @@ var _ = Describe("Probe [Identity Service]", func() {
 		Expect(ok).To(BeTrue())
 		Expect(serverError.Code() == codes.FailedPrecondition ||
 			serverError.Code() == codes.OK).To(BeTrue())
+
+		if res.GetReady() != nil {
+			Expect(res.GetReady().GetValue() == true ||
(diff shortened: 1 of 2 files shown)


pkg/sanity/identity.go
```diff
@@ -79,6 +79,8 @@ var _ = Describe("Probe [Identity Service]", func() {
 		Expect(ok).To(BeTrue())
 		Expect(serverError.Code() == codes.FailedPrecondition ||
 			serverError.Code() == codes.OK).To(BeTrue())
+
+		Expect(res.GetReady().GetValue()).To(BeTrue())
```
darkowlzz: I'm not sure if we should expect this to be true always, but if it's not true everything might fail.

lpabon: Couple of things.

1. The value can be `true` or `false`. We should not make sure it is true, just make sure it is true or false.
2. We need to check first if the value is set since according to the spec, this is optional. So first we need to

```go
if req.GetRead() != nil {
    // Check the values 
}
```

darkowlzz: Thanks! That cleared my confusion :)

darkowlzz:
Added the requested changes in the same commit. Hope this is ready now :)

PR #80, opened 2018-06-17
sanity: Add ListVolumes test to check starting token and next token

darkowlzz:
- Adds test for expecting `ABORTED` gRPC code when an invalid
starting_token is passed to `Controller.ListVolumes()` and when the
starting_token value is greater than the actual number of volumes.

- Adds a `Controller.ListVolumes()` test to check for the
change in volume list when a volume is added and removed.

- Adds `Controller.ListVolumes()` test to pass `MaxEntries` and receive a
`NextToken` and ensure paginated list volumes works.

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -19,13 +19,12 @@ package sanity
 import (
 	"context"
 	"fmt"
+	"strconv"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
-
-	"strconv"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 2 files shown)


pkg/sanity/controller.go
```diff
@@ -193,8 +193,73 @@ var _ = Describe("ListVolumes [Controller Server]", func() {
 		Expect(serverError.Code()).To(Equal(codes.Aborted))
 	})
 
-	// TODO: Add test which checks list of volume is there when created,
-	//       and not there when deleted.
+	It("check the presence of new volumes in the volume list", func() {
```
darkowlzz: >  // TODO: Add test which checks list of volume is there when created, 
>  //       and not there when deleted. 

It wasn't clear what this TODO was actually about, but I felt it's about adding and removing, and checking volume list. Need a little verification.

pkg/sanity/controller.go
```diff
@@ -152,10 +153,213 @@ var _ = Describe("ListVolumes [Controller Server]", func() {
...
+		)
+		Expect(err).To(HaveOccurred())
+		Expect(vols).To(BeNil())
+
+		serverError, ok := status.FromError(err)
+		Expect(ok).To(BeTrue())
+		Expect(serverError.Code()).To(Equal(codes.Aborted))
+	})
+
+	It("check the presence of new volumes in the volume list", func() {
```
darkowlzz: NOTE: _Adding this comment here again, because the previous commit appears to be outdated by the latest commit, making the comment in the previous commit be folded/hidden._

>  // TODO: Add test which checks list of volume is there when created, 
>  //       and not there when deleted. 

It wasn't clear what this TODO was actually about, but I felt it's about adding and removing, and checking volume list. Need a little verification.

pkg/sanity/controller.go
```diff
@@ -152,10 +153,213 @@ var _ = Describe("ListVolumes [Controller Server]", func() {
...
+		vols, err = c.ListVolumes(
+			context.Background(),
+			&csi.ListVolumesRequest{
+				StartingToken: nextToken,
+			})
+		Expect(err).NotTo(HaveOccurred())
+		Expect(vols).NotTo(BeNil())
+
+		// Ensure that all the remaining entries are returned at once.
+		Expect(len(vols.GetEntries())).To(Equal(currentTotalVols - maxEntries))
```
lpabon: Interesting, this could be left to interpretation, but the driver could have divided the list into segments equal to the initial List(MaxEntries) size. 

Here is my suggestion. For now, set the test to be Skipped so that we do not loose it, and the go to the CSI team and confirm your interpretation of the CSI ListVolumes request. Make sure that is correct. You can also just send an update PR to the comment of ListVolumes and discuss it there.

pkg/sanity/controller.go
```diff
@@ -152,10 +153,213 @@ var _ = Describe("ListVolumes [Controller Server]", func() {
...
+
+		if initialTotalVols < minVolCount {
+
+			By("cleaning up deleting the volumes")
+			requiredVols := minVolCount - initialTotalVols
+			name := "sanity"
+			for i := 1; i <= requiredVols; i++ {
+				delReq := &csi.DeleteVolumeRequest{
+					VolumeId: name + strconv.Itoa(i),
```
wackxu: I think this is not right, `name + strconv.Itoa(i)`  is the volume name that we create, not the id, I think we can save the volume that we create in a slice, then we can use the slice info to delete the volume we create.

lpabon:
@darkowlzz Any updates?

pkg/sanity/controller.go
darkowlzz: Thanks!

pkg/sanity/controller.go
darkowlzz: Sorry for the delay.
Reading the spec again, 
> If not specified (zero value), it means there is no restriction on the number of entries that can be returned.

Sounds like we need not check if all the remaining entries are returned. The test is about testing if a next token is returned. So, it's better to remove this part of the test only.

PR #83, opened 2018-06-20
add test for volume snapshot

wackxu:
/assign @lpabon

### DIFF
--- a/mock/service/controller.go
+++ b/mock/service/controller.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math"
 	"path"
+	"reflect"
 	"strconv"
 
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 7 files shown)


wackxu:
@darkowlzz Thanks for your review,  will update it.

mock/service/service.go
```diff
@@ -109,3 +123,53 @@ func (s *service) findVolByName(
...
+	}
+
+	return
+}
+
+func (s *service) findSnapshotByName(
+	ctx context.Context, name string) (int, csi.Snapshot) {
+
+	return s.findSnapshot("name", name)
+}
```
darkowlzz: Let's add a newline at the end.

pkg/sanity/controller.go
```diff
@@ -48,6 +48,11 @@ func verifyVolumeInfo(v *csi.Volume) {
 	Expect(v.GetId()).NotTo(BeEmpty())
 }
 
+func verifySnapshotInfo(snapshot *csi.Snapshot) {
+	Expect(snapshot).NotTo(BeNil())
+	Expect(snapshot.GetId()).NotTo(BeEmpty())
```
darkowlzz: I think we can add some more verification here.

`SourceVolumeID` is REQUIRED.

Snapshot definition talks about `SizeBytes`.
> The size of the volume MUST NOT  be less than the size of the source snapshot.

Let's add those.

pkg/sanity/controller.go
```diff
@@ -933,3 +938,382 @@ var _ = Describe("ControllerUnpublishVolume [Controller Server]", func() {
...
+
+		for _, snapshot := range snapshots.GetEntries() {
+			verifySnapshotInfo(snapshot.GetSnapshot())
+		}
+	})
+
+	// TODO: Add test to test for tokens
+
+	// TODO: Add test which checks list of snapshot is there when created,
+	//       and not there when deleted.
```
darkowlzz: I've added some tests in `ListVolumes` which are similar to what we need here. Maybe have a look and add them? https://github.com/kubernetes-csi/csi-test/pull/80
Or maybe as a follow-up PR?

pkg/sanity/sanity.go
```diff
@@ -40,6 +40,8 @@ type CSISecrets struct {
 	ControllerUnpublishVolumeSecret map[string]string `yaml:"ControllerUnpublishVolumeSecret"`
 	NodeStageVolumeSecret           map[string]string `yaml:"NodeStageVolumeSecret"`
 	NodePublishVolumeSecret         map[string]string `yaml:"NodePublishVolumeSecret"`
+	CreateSnapshotSecret            map[string]string `yaml:"CreateSnapshotSecret"`
+	DeleteSnapshotSecret            map[string]string `yaml:"DeleteSnapshotSecret"`
```
darkowlzz: Along with these for snapshot secrets in the mock driver, you should also add snapshot secret authentication in driver/driver.go, like https://github.com/kubernetes-csi/csi-test/blob/master/driver/driver.go#L198-L204.
And then some snapshot secrets in https://github.com/kubernetes-csi/csi-test/blob/master/mock/mocksecret.yaml. Those would be used by the mock test run with secrets enabled,

mock/service/controller.go
```diff
@@ -338,21 +338,183 @@ func (s *service) ControllerGetCapabilities(
...
 	req *csi.ListSnapshotsRequest) (*csi.ListSnapshotsResponse, error) {
-	return nil, status.Error(codes.InvalidArgument, "Not Implemented")
+
+	// case 1: SnapshotId is not empty, return snapshots that match the snapshot id.
+	if len(req.GetSnapshotId()) != 0 {
+		i, snapshot := s.findSnapshotNoLock("id", req.SnapshotId)
+		if i < 0 {
+			return &csi.ListSnapshotsResponse{}, nil
+		}
+		return &csi.ListSnapshotsResponse{
```
darkowlzz: As per the spec:
> ListSnapshots SHALL NOT list a snapshot that is being created but has not been cut successfully yet.

So, let's check if `snapshot.GetStatus().Type` is `SnapshotStatus_READY` before adding it into the list of snapshots.

mock/service/controller.go
```diff
@@ -338,21 +338,183 @@ func (s *service) ControllerGetCapabilities(
...
+			Entries: []*csi.ListSnapshotsResponse_Entry{
+				{
+					Snapshot: &snapshot,
+				},
+			},
+		}, nil
+	}
+
+	// case 2: SourceVolumeId is not empty, return snapshots that match the source volume id.
+	if len(req.SourceVolumeId) != 0 {
```
darkowlzz: Would recommend using `req.GetSourceVolumeId()`.

mock/service/controller.go
```diff
@@ -338,21 +338,183 @@ func (s *service) ControllerGetCapabilities(
...
+			return &csi.ListSnapshotsResponse{}, nil
+		}
+		return &csi.ListSnapshotsResponse{
+			Entries: []*csi.ListSnapshotsResponse_Entry{
+				{
+					Snapshot: &snapshot,
+				},
+			},
+		}, nil
+	}
```
darkowlzz: I think we need to consider both `req.GetSnapshotId()` and `req.GetSourceVolumeID` when they aren't empty. If the volume received from ID has a different source volume ID, do not include them in the snapshot list.

mock/service/controller.go
```diff
@@ -338,21 +338,183 @@ func (s *service) ControllerGetCapabilities(
...
+
+	// case 2: SourceVolumeId is not empty, return snapshots that match the source volume id.
+	if len(req.SourceVolumeId) != 0 {
+		i, snapshot := s.findSnapshotNoLock("id", req.SourceVolumeId)
```
darkowlzz: `"id"` -> `"sourceVolumeId"`

.idea/workspace.xml
```diff
@@ -0,0 +1,499 @@
...
+    <entry file="file://$PROJECT_DIR$/mock/service/controller.go">
+      <provider selected="true" editor-type-id="text-editor">
+        <state relative-caret-position="11178">
+          <caret line="516" column="27" lean-forward="false" selection-start-line="516" selection-start-column="27" selection-end-line="516" selection-end-column="27" />
+          <folding />
+        </state>
+      </provider>
+    </entry>
+  </component>
+</project>
```
darkowlzz: I don't think we need this file :)

mock/service/controller.go
```diff
@@ -271,7 +271,7 @@ func (s *service) ListVolumes(
 		i       int
 		j       = startingToken
 		entries = make(
-			[]*csi.ListVolumesResponse_Entry,
+		[]*csi.ListVolumesResponse_Entry,
```
darkowlzz: Looked better before :)

.idea/workspace.xml
lpabon: Yes, please remove this file.

mock/service/controller.go
```diff
@@ -338,21 +338,183 @@ func (s *service) ControllerGetCapabilities(
...
+	if len(req.GetSourceVolumeId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "SourceVolumeId missing in request")
+	}
+
+	// Check to see if the snapshot already exists.
+	if i, v := s.findSnapshotByName(ctx, req.Name); i >= 0 {
+		// Requested snapshot name already exists
+		if v.GetSourceVolumeId() == req.SourceVolumeId {
+			return nil, status.Error(codes.AlreadyExists,
+				fmt.Sprintf("Snapshot with name %s already exists", req.GetName()))
```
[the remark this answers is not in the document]
xing-yang: +1. If source_volume_id, create_snapshot_secrets, and parameters are the same for the same snapshot name, it should return OK for idempotency.

lpabon:
Seems that there are two CI errors: Go fmt errors and some failed csi-sanity errors

mock/service/controller.go
```diff
@@ -338,21 +338,183 @@ func (s *service) ControllerGetCapabilities(
...
 		},
 	}, nil
 }
 
 func (s *service) CreateSnapshot(ctx context.Context,
 	req *csi.CreateSnapshotRequest) (*csi.CreateSnapshotResponse, error) {
-	return nil, status.Error(codes.InvalidArgument, "Not Implemented")
+	// Check arguments
+	if len(req.GetName()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "Name missing in request")
```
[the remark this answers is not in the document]
wackxu: Done

mock/service/controller.go
```diff
@@ -338,21 +338,183 @@ func (s *service) ControllerGetCapabilities(
...
+	}
+	if len(req.GetSourceVolumeId()) == 0 {
+		return nil, status.Error(codes.InvalidArgument, "SourceVolumeId missing in request")
```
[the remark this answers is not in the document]
wackxu: Done

mock/service/controller.go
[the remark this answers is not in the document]
wackxu: Updated for check parameters also, I do not think we should check create_snapshot_secrets that may change every request.

mock/service/service.go
wackxu: Done

pkg/sanity/controller.go
wackxu: Updated for check attributes  that is REQUIRED. As for `SizeBytes`, 
>The purpose of this field is to give CO guidance on how much space is needed to
	// create a volume from this snapshot. The size of the volume MUST NOT
	// be less than the size of the source snapshot.

I think it is used to CO when restore volume from a snapshot to verify there is enough space. So we do not need verify here.

pkg/sanity/sanity.go
wackxu: Thanks for your help, Updated.

mock/service/controller.go
wackxu: Done

mock/service/controller.go
wackxu: Done

mock/service/controller.go
wackxu: Thanks, done

mock/service/controller.go
wackxu: Done

wackxu:
@lpabon @darkowlzz updated all comments, PTAL

mock/service/controller.go
```diff
@@ -367,21 +368,193 @@ func (s *service) ControllerGetCapabilities(
...
+	if i < 0 {
+		return &csi.DeleteSnapshotResponse{}, nil
+	}
+
+	// This delete logic preserves order and prevents potential memory
+	// leaks. The slice's elements may not be pointers, but the structs
+	// themselves have fields that are.
+	copy(s.snapshots[i:], s.snapshots[i+1:])
+	s.snapshots[len(s.snapshots)-1] = Snapshot{}
+	s.snapshots = s.snapshots[:len(s.snapshots)-1]
```
lpabon: line 440 removes the last element which was just initialized in line 439. It looks to me that line 439 is not needed, right?

mock/service/controller.go
```diff
@@ -367,21 +368,193 @@ func (s *service) ControllerGetCapabilities(
...
+		}, nil
+	}
+
+	// case 3: no parameter is set, so we return all the snapshots.
+
+	// Copy the mock snapshots into a new slice in order to avoid
+	// locking the service's snapshot slice for the duration of the
+	// ListSnapshots RPC.
+	var snapshots []csi.Snapshot
+	func() {
```
lpabon: I've never seen an anonymous function without a `go` statement or assigned to a variable. Just curious, but why use this style?

lpabon: FYI, when I review, I'm not only looking for bugs, but I am looking at maintenance. If it is hard to maintain, even if it works, it should be redone.

mock/service/controller.go
```diff
@@ -367,21 +368,193 @@ func (s *service) ControllerGetCapabilities(
...
+		s.snapshotsRWL.RLock()
```
lpabon: Have to be very careful creating a deadlock situation sometime in the future. I would instead place these locks as close to the object as possible. I would suggest creating a new data structure object that handles the the work with the list. It supports Insert, Delete, List, Find, etc. and it is the one managing the lock

mock/service/controller.go
```diff
@@ -367,21 +368,193 @@ func (s *service) ControllerGetCapabilities(
...
+	s.snapshots = s.snapshots[:len(s.snapshots)-1]
+	log.WithField("SnapshotId", req.SnapshotId).Debug("mock delete snapshot")
+	return &csi.DeleteSnapshotResponse{}, nil
 }
 
 func (s *service) ListSnapshots(ctx context.Context,
 	req *csi.ListSnapshotsRequest) (*csi.ListSnapshotsResponse, error) {
-	return nil, status.Error(codes.InvalidArgument, "Not Implemented")
+
+	// case 1: SnapshotId is not empty, return snapshots that match the snapshot id.
```
lpabon: This function is too large, do you mind breaking it up by the cases you have?

mock/service/controller.go
wackxu: Just imitate the code about volume deletion, make sense, will delete it.

mock/service/controller.go
wackxu: Just imitate the code about volume deletion, make sense, I will update the code and commit a separate pr to change the volume part.

mock/service/controller.go
wackxu: Make sense, will update.

PR #84, opened 2018-06-22
Add test for ControllerPublishVolume

codenrhoden:
I'm opening this PR to demonstrate an error I am seeing with `csi-test` master. When updating the CSI external attacher/detacher, I always get a test failure for ControllerPublishVolume. In fact it hangs - took a while to track it down to this.

The `reflect.DeepEqual` that `gomock` performs to check for object equivalence is failing for the `ControllerPublishVolume` input. Oddly, it is *only* that call that fails - all the other tests pass. For simplicity, I added a test here in `csi-test` that demonstrates it.

When the test fails, you will see the following:
```
Unexpected call to *driver.MockControllerServer.ControllerPublishVolume([context.Background.WithCancel.WithCancel.WithValue(peer.peerKey{}, &peer.Peer{Addr:(*net.TCPAddr)(0xc4201b0c60), AuthInfo:credentials.AuthInfo(nil)}).WithValue(metadata.mdIncomingKey{}, metadata.MD{":authority":[]string{"127.0.0.1:51284"}, "content-type":[]string{"application/grpc"}, "user-agent":[]string{"grpc-go/1.12.2"}}).WithValue(grpc.streamKey{}, <stream: 0xc42032a000, /csi.v0.Controller/ControllerPublishVolume>) volume_id:"myname" node_id:"MyNodeID" volume_capability:<mount:<> access_mode:<mode:MULTI_NODE_MULTI_WRITER > > ]) at /Users/trhoden/go/src/github.com/kubernetes-csi/csi-test/driver/driver.mock.go:522 because:
Expected call at /Users/trhoden/go/src/github.com/kubernetes-csi/csi-test/driver/driver.mock.go:529 doesn't match the argument at index 1.
Got: volume_id:"myname" node_id:"MyNodeID" volume_capability:<mount:<> access_mode:<mode:MULTI_NODE_MULTI_WRITER > >
Want: is equal to volume_id:"myname" node_id:"MyNodeID" volume_capability:<mount:<> access_mode:<mode:MULTI_NODE_MULTI_WRITER > >
panic: MOCK TEST FATAL FAILURE
```

Note that the Expected and Got lines show the object is correct, but it is not being picked up that way. I've tried using different versions of `gomock`, `grpc`, regenerating the mock objects, etc., but haven't gotten around this one yet. Help is appreciated!

### DIFF
--- a/test/co_test.go
+++ b/test/co_test.go
@@ -16,10 +16,13 @@ limitations under the License.
 package test
 
 import (
+	"fmt"
+	"reflect"
 	"testing"
 
 	csi "github.com/container-storage-interface/spec/lib/go/csi/v0"
 	gomock "github.com/golang/mock/gomock"
+	"github.com/golang/protobuf/proto"
 	mock_driver "github.com/kubernetes-csi/csi-test/driver"
 	mock_utils "github.com/kubernetes-csi/csi-test/utils"
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


codenrhoden:
I identified the underlying issue to be that since the switch to protobuf 1.1.1, one cannot use `reflect.DeepEqual` to test for equivalence of `proto.Message`. So, I added a custom matcher that instead uses `proto.Equal`, and things look good!

See https://groups.google.com/forum/#!topic/protobuf/N-elvFu4dFM for details on the protobuf change. This should be ready to merge now.

lpabon:
@codenrhoden Hi, thanks for the PR. What do you suggest. Would you suggest adding this as a library to ci-test?

PR #90, opened 2018-07-05
pkg/sanity: support sanity testing in another Ginkgo suite

pohly:
This was not possible before:
- Ginkgo only supports one set of Before/AfterSuite calls,
  which might be in use already in the existing suite (for
  example, the Kubernetes e2e framework).
- When nesting the sanity test inside a Ginkgo spec there
  is no testing.T pointer that could be passed to
  the Test method.

Now all global variables are replaced with a test context that gets
passed into the specs explicitly. For this to work, specs must be set
up with SanityDescribe instead of the previous Describe call.

The Test and Config API are unchanged, but there is one slight change
of behavior: previously it was possible to define two tests with
different configs in the same test binary and then running each test
would execute the sanity tests for that config. That no longer works
because running one test changes the global Ginkgo suite and running
the second test would then also run the specs from the first one.

This usage could be supported by having one test function which takes
multiple configs, then runs one suite that contains tests for all of
them.

Probably it is easier to just use a Ginkgo suite, because there
multiple different configs can be handled already normally via
different Describe specs: calling the new GinkgoTest API will add all
sanity test cases for one configuration to the spec in which the API
is called.

### DIFF
--- a/pkg/sanity/README.md
+++ b/pkg/sanity/README.md
@@ -13,12 +13,49 @@ Golang `TestXXX` functions. For example:
 
 ```go
 func TestMyDriver(t *testing.T) {
-    // Setup the full driver and its environment
-    ... setup driver ...
+	// Setup the full driver and its environment
+	... setup driver ...
+	config := &sanity.Config{
+		TargetPath:     ...
+		StagingPath:    ...
+		Address:        endpoint,
+	}
 
-    // Now call the test suite
-    sanity.Test(t, driverEndpointAddress, "/mnt")
+
+	// Now call the test suite
+	sanity.Test(t, config)
 }
+```
+
+Only one such test function is supported because under the hood a
+Ginkgo test suite gets constructed and executed by the call.
+
+Alternatively, the tests can also be embedded inside a Ginkgo test
+suite. In that case it is possible to define multiple tests with
+different configurations:
+
+```go
+var _ = Describe("MyCSIDriver", func () {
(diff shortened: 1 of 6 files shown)


lpabon:
Hi @pohly , this vet and unit test issues.

pohly:
Ah, those are in .travis.yml. I was looking for a "make check" or "make test" and missed the .travis.yml. I'll fix these issues.

pohly:
`make build` doesn't rebuild when the sources change. Should we make $(APP) a phony target to ensure that `go build` is always invoked?

lpabon:
>make build doesn't rebuild when the sources change. Should we make $(APP) a phony target to ensure that go build is always invoked? And should the top-level Makefile also build the csi-sanity command? Currently one has to remember to invoke "make" in two different directories.

If you want to, please feel free to send a PR for this.

pohly:

>>make build doesn't rebuild when the sources change. Should we make $(APP) a phony target to ensure that go build is always invoked? And should the top-level Makefile also build the csi-sanity command? Currently one has to remember to invoke "make" in two different directories.
>
> If you want to, please feel free to send a PR for this.

@lpabon I ended up adding this to my pending PR #93 because I wanted to
simplify testing that while I worked on it, please see
https://github.com/kubernetes-csi/csi-test/pull/93/commits/e2e872d7fb9497e2f83089cc50cad379b98487af.

Issue #89, opened 2018-07-04
Sanity tests may not always cleanup


msau42:
There are two problems with the tests:
1. The CreateVolume request could succeed, but subsequent validation could fail.  DeleteVolume() is skipped in those cases so the volume never gets cleaned up.
2. Almost all the tests call CreateVolume with volume id "sanity".  It would be better to use unique names for every test case to avoid potential false positives and issues where a previously failing test causes a subsequent test to fail because the volume already exists.

pohly:
Reusing the same volume ID also prevents running specs in parallel. I'm encountering that right now.

I'll work on a fix.

pohly:
@msau42 should each test *case* have a unique name or also each test *run*? In other words, is is okay to have fixed names like `sanity-controller-single-with-capacity` (unique for each test case) or do we need random names like `sanity-44de2d24-7f9a-11e8-8a87-77571b0768c3`?

I prefer `sanity-controller-single-with-capacity`. It makes it easier to follow and repeat tests, but of course the downside is that one might have to clean up after an aborted test run which left volumes behind.

pohly:
Does it make sense to run the exact same test twice, under different names?

See NodeStageVolume and NodeUnstageVolume - both run testFullWorkflowSuccess.

pohly:
Not twice - four times!

msau42:
If the driver is persisting volume names to disk, then the issue could still exist.  Mine only keeps the names in memory, so it's fine for me.

I think the tests could also be restructured a bit to add the volume cleanup to an aftereach so that it will always happen even if the test errors early.

pohly:
Okay, so let's use a unique suffix. I'll also add the cleanup and restructure the tests so that the testFullWorkflowSuccess test only runs once.

PR #93, opened 2018-07-11
test suite reorganization

pohly:
This fixes issue #89 (deleting volumes on test failure, unique volume names) and simplifies the code a bit.

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -167,14 +167,12 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 		})
 
 		It("should fail when no name is provided", func() {
-
-			req := &csi.CreateVolumeRequest{}
-
-			if sc.Secrets != nil {
-				req.ControllerCreateSecrets = sc.Secrets.CreateVolumeSecret
-			}
-
-			_, err := c.CreateVolume(context.Background(), req)
+			_, err := c.CreateVolume(
+				context.Background(),
+				&csi.CreateVolumeRequest{
+					ControllerCreateSecrets: sc.Secrets.CreateVolumeSecret,
+				},
+			)
 			Expect(err).To(HaveOccurred())
 
 			serverError, ok := status.FromError(err)
@@ -183,16 +181,13 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 		})
 
 		It("should fail when no volume capabilities are provided", func() {
-
-			req := &csi.CreateVolumeRequest{
-				Name: "name",
-			}
-
-			if sc.Secrets != nil {
-				req.ControllerCreateSecrets = sc.Secrets.CreateVolumeSecret
-			}
-
-			_, err := c.CreateVolume(context.Background(), req)
+			_, err := c.CreateVolume(
+				context.Background(),
+				&csi.CreateVolumeRequest{
+					Name: "name",
+					ControllerCreateSecrets: sc.Secrets.CreateVolumeSecret,
+				},
+			)
 			Expect(err).To(HaveOccurred())
 
 			serverError, ok := status.FromError(err)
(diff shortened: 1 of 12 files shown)


pkg/sanity/sanity.go
```diff
@@ -136,3 +137,23 @@ func loadSecrets(path string) (*CSISecrets, error) {
...
 	return &creds, nil
 }
+
+var uniqueSuffix = "-" + pseudoUUID()
+
+// pseudoUUID returns a unique string generated from random
+// bytes, empty string in case of error.
+func pseudoUUID() string {
+	b := make([]byte, 8)
+	if _, err := rand.Read(b); err != nil {
```
msau42: Do you need to seed the random generator at the beginning?

pkg/sanity/controller.go
```diff
@@ -265,6 +279,7 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 				Expect(vol.GetVolume().GetId()).NotTo(BeEmpty())
 				Expect(vol.GetVolume().GetCapacityBytes()).To(BeNumerically(">=", TestVolumeSize(sc)))
 			}
+			cl.RegisterVolume(name, VolumeInfo{VolumeID: vol.GetVolume().GetId()})
```
msau42: Should this go after L279

pkg/sanity/controller.go
```diff
@@ -210,6 +222,7 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 				},
 			)
 			Expect(err).NotTo(HaveOccurred())
+			cl.RegisterVolume(name, VolumeInfo{VolumeID: vol.GetVolume().GetId()})
```
msau42: Should this go after L228 because it requires vol.GetVolume().GetId()?

pkg/sanity/controller.go
```diff
@@ -210,6 +222,7 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
...
 			Expect(vol).NotTo(BeNil())
```
msau42: All of the test cases are validating these same things. Can the validation be put into a common function?

pkg/sanity/sanity.go
pohly: This is `crypto/rand`, which doesn't need to be seeded.

One could argue that the normal `rand` is good enough, but then it would need to be seeded. I don't care either way.

pkg/sanity/controller.go
pohly: I was wondering the same about the full test and ended up moving that to a single test and then again here after making the same change over and over again.

I guess some more tests could be eliminated, but I haven't looked too closely and would prefer to get this PR merged first before doing further simplifications.

pkg/sanity/controller.go
pohly: Yes, will change that.

pkg/sanity/controller.go
pohly: Yes, will change that.

pkg/sanity/controller.go
msau42: I don't mean eliminating tests, but consolidating the 4 checks about create volume return values into a common function

pohly: The downside of putting the checks into a common function is that line information in case of a failure isn't useful anymore:
```
• Failure [0.017 seconds]
Controller Service
/fast/work/gopath/src/github.com/intel/open-infrastructure-offload/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
  CreateVolume
  /fast/work/gopath/src/github.com/intel/open-infrastructure-offload/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:173
    should return appropriate values SingleNodeWriter WithCapacity 1Gi Type:Mount [It]
    /fast/work/gopath/src/github.com/intel/open-infrastructure-offload/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:250

    Expected an error to have occurred.  Got:
        <nil>: nil

    /fast/work/gopath/src/github.com/intel/open-infrastructure-offload/vendor/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:73
```

This line 73 is in a new helper function where I intentionally inverted the error check to trigger a test failure.

I think this needs further discussion, so I'll leave it as it is.

pohly:
@msau42 I've force-pushed a version where the `RegisterVolume` calls are after verifying the volume ID. The odd control flow around the OutOfRange error detection confused me, so I simplified that a bit. Please have another look.

pkg/sanity/controller.go
```diff
@@ -69,1100 +69,1014 @@ func isControllerCapabilitySupported(
...
-					AccessMode: &csi.VolumeCapability_AccessMode{
-						Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+		It("should fail when no name is provided", func() {
+			_, err := c.CreateVolume(
+				context.Background(),
+				&csi.CreateVolumeRequest{
+					ControllerCreateSecrets: sc.Secrets.CreateVolumeSecret,
+				},
+			)
+			Expect(err).To(HaveOccurred())
```
msau42: If CreateVolume actually succeeded, then you still need to cleanup the volume.  Same for all negative tests.

pohly: Yes, I had asked myself about that too, but then dismissed it as unlikely. Will add it.

pohly: Done, see d34fca7.

pkg/sanity/controller.go
```diff
@@ -171,12 +171,13 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
...
 
 		It("should fail when no name is provided", func() {
-			_, err := c.CreateVolume(
+			vol, err := c.CreateVolume(
 				context.Background(),
 				&csi.CreateVolumeRequest{
 					ControllerCreateSecrets: sc.Secrets.CreateVolumeSecret,
 				},
 			)
+			cl.MaybeRegisterVolume("", vol, err)
```
msau42: This volume doesn't have a name?

lpabon:
@pohly Could you squash your changes when you resolve the conflict?

pohly:
> @pohly Could you squash your changes when you resolve the conflict?

Okay, done.

Please note that I included another fix that wasn't part of the PR before:

275a08636f8 - "Makefile: fix format error reporting"

msau42:
@pohly @lpabon can we get this merged?

pohly:
It's ready as far as I'm concerned. @lpabon, please merge.

PR #97, opened 2018-08-03
utils: Add grpc middleware for connection retry

darkowlzz:
This adds a grpc client dialOption for retrying when a connection times
out. With this the client would retry a few times before failing
immediately after a timeout.

**Issue Description**

I've been observing connection timeouts after upgrading from v0.2.0 to v0.3.0. And they happen at random parts of the test suite. Here's an example when it happened for `NodeUnpublishVolume`:
```
STEP: connecting to CSI driver
[AfterEach] NodeUnpublishVolume [Node Server]
 /home/lpabon/git/golang/csi-test/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:50

• Failure in Spec Setup (BeforeEach) [60.007 seconds]
NodeUnpublishVolume [Node Server]
/home/lpabon/git/golang/csi-test/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
 should fail when no target path is provided [BeforeEach]
 /home/lpabon/git/golang/csi-test/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:263

 Expected error:
     <*errors.errorString | 0xc420332410>: {
         s: "Connection timed out",
     }
     Connection timed out
 not to have occurred
```
I've seen this happening for other cases as well.
I thought it might be something to do with my driver, but when I run v0.2.0 right after v0.3.0 fails, no failures for v0.2.0.
Running v0.2.0 successively never fails, but running v0.3.0 successively results in frequent failures.

Adding retry capability in the grpc client seemed to have worked in avoiding the failures. This PR reduces the connection WaitForStateChange time and retries a few times before returning a timeout error.


cmd/csi-sanity/Makefile
```diff
@@ -1,5 +1,5 @@
 APP_NAME := csi-sanity
-VER :=$(shell git describe)
+VER :=$(shell git describe --tags --dirty)
```
darkowlzz: Due to some reason, just `describe` was returning v0.2.0 tag for me. After setting `--tags`, it returned v0.3.0. And `--dirty` is nice to have :)
Hope it's fine to have these.

utils/grpcutil.go
```diff
@@ -41,16 +44,25 @@ func Connect(address string) (*grpc.ClientConn, error) {
...
+	// Add retry support.
+	dialOptions = append(dialOptions, grpc.WithUnaryInterceptor(grpc_retry.UnaryClientInterceptor()))
+
 	conn, err := grpc.Dial(address, dialOptions...)
 	if err != nil {
 		return nil, err
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
darkowlzz: I'm not sure if we really need to wait for a minute. But maybe 10 seconds would be too short. Would need help here.

lpabon:
Do not merge.. @darkowlzz is investigating the root cause.

darkowlzz:
Investigation report :)

I did git-bisect and found that this behavior was introduced in https://github.com/kubernetes-csi/csi-test/commit/44a5850955843e5a8293e0f7a2556a7bdca8d5a6 . On examining the changes, it appeared that before that change, we used to have a single connection to the CSI driver and use it for all the tests. Since that commit, we started storing the grpc client connection in `SanityContext` object, which gets created for every test case. This resulted in multiple connection creation and deletion.

I've seen this connection timeout while running the e2e tests with the mock driver as well:
```
$ ./hack/e2e.sh 
rm -f csi-sanity
go test -ldflags "-w -X github.com/kubernetes-csi/csi-test/cmd/csi-sanity.VERSION=v0.3.0-1-13-gf272f52-dirty -extldflags '-z relro -z now'" -c -o csi-sanity
cp csi-sanity /home/darkowlzz/go/bin
mock driver started
Running Suite: CSI Driver Test Suite
====================================
Random Seed: 1533374841
Will run 58 of 58 specs

•••••••STEP: connecting to CSI driver

------------------------------
• Failure in Spec Setup (BeforeEach) [60.003 seconds]
Node Service
/home/darkowlzz/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
  NodePublishVolume [BeforeEach]
  /home/darkowlzz/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:168
    should fail when no target path is provided
    /home/darkowlzz/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:183

    Expected error:
        <*errors.errorString | 0xc4204dc0a0>: {
            s: "Connection timed out",
        }
        Connection timed out
    not to have occurred

    /home/darkowlzz/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/sanity.go:97
------------------------------
••••••••••••••••••••••••••••••••••••••••••••••••••

Summarizing 1 Failure:

[Fail] Node Service [BeforeEach] NodePublishVolume should fail when no target path is provided 
/home/darkowlzz/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/sanity.go:97

Ran 58 of 58 Specs in 60.109 seconds
FAIL! -- 57 Passed | 1 Failed | 0 Pending | 0 Skipped
--- FAIL: TestSanity (60.12s)
FAIL

```

If we continue creating multiple connections, we can use this PR to attempt retry when there's a connection timeout. Or, I've made some changes (another PR) to have a single global connection object, like before, which is reused by all the test cases. The whole test suite seems to be faster and reliable with a single connection.

lpabon:
Closing in favor of #105

PR #98, opened 2018-08-04
pkg/sanity: Use a single grpc connection

darkowlzz:
This change makes the grpc client connection to be stored as a global
variable and used across all the test cases. This helps avoid creating
multiple connections at every setup and teardown cycle. The whole test
suite runs faster and more reliable with a single shared connection.

Alternate solution to #97

### DIFF
--- a/pkg/sanity/sanity.go
+++ b/pkg/sanity/sanity.go
@@ -95,6 +95,7 @@ func GinkgoTest(reqConfig *Config) {
 		Config: reqConfig,
 	}
 
+	address = sc.Config.Address
 	registerTestsInGinkgo(sc)
 }
 
--- a/pkg/sanity/sanity.go
+++ b/pkg/sanity/sanity.go
@@ -58,9 +58,13 @@ type Config struct {
 // gets initialized before each test block runs.
 type SanityContext struct {
 	Config  *Config
-	Conn    *grpc.ClientConn
 	Secrets *CSISecrets
 }
+
+var (
+	conn    *grpc.ClientConn
+	address string
+)
 
 // Test will test the CSI driver at the specified address by
 // setting up a Ginkgo suite and running it.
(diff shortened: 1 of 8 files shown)


darkowlzz:
I felt that the changes in this PR broke the api and ginkgo embedded use case. Added two separate tests for those. I can move these two tests to #97 if this solution has some drawbacks.

pkg/sanity/sanity.go
```diff
@@ -58,27 +58,44 @@ type Config struct {
...
 		Config: reqConfig,
 	}
 
+	address = sc.Config.Address
 	registerTestsInGinkgo(sc)
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "CSI Driver Test Suite")
 }
 
+var _ = BeforeSuite(func() {
```
lpabon: I think @pohly 's change required these not to be set. Please check with him over slack.

darkowlzz: Yes, now it makes sense why this was removed. Embedding this test in an existing ginkgo setup with existing global `BeforeSuite` and `AfterSuite` fails to run. Maybe it's better to go with connection retry approach. The examples in ginkgo docs create multiple DB connections for each test case to avoid test pollution.

I don't have much experience with ginkgo. @pohly what do you think? Maybe there's some other better way to fix this?

pohly:
Sunny <notifications@github.com> writes:

> darkowlzz commented on this pull request.
>
>
>
>>  	registerTestsInGinkgo(sc)
>  	RegisterFailHandler(Fail)
>  	RunSpecs(t, "CSI Driver Test Suite")
>  }
>  
> +var _ = BeforeSuite(func() {
>
> Yes, now it makes sense why this was removed. Embedding this test in
> an existing ginkgo setup with existing global `BeforeSuite` and
> `AfterSuite` fails to run. Maybe it's better to go with connection
> retry approach.

I think this would be the cleaner approach. Clients of a CSI driver must
be prepared for the CSI driver to go down and come up again and should
be able to handle the resulting connection issues.

I'm wondering whether current or future gRPC releases will handle this
also for Unix domain sockets automatically. I wanted to investigate
this, but haven't gotten around to it yet, and currently I am on
vacation. I found the workaround with a custom dialer for Unix domain
sockets rather odd because the gRPC spec has unix:/// as one of the
methods, but then I learned that Go gRPC itself (at least in older
versions) didn't actually support that.

> I don't have much experience with ginkgo. @pohly what do you think?
> Maybe there's some other better way to fix this?

The alternative is to set up a global connection on demand and never
close it. Then all tests running in that Ginkgo worker process (usually
one, unless parallelism is explicitly enabled) share the same
connection.

lpabon:
closing in favor of #105

PR #102, opened 2018-08-14
Update Gopkg.lock for dep 0.5.0

darkowlzz:
The new version of dep adds some extra information in the lock file for
vendor verification.


pohly:
We'e already switch to dep 0.5.0 on master.

/close

PR #105, opened 2018-08-30
pkg/sanity: Use a single grpc connection via caching in SanityContext

pohly:
This uses the same approach as PR #98 to combat connection issues (connecting only once per process), but implements it so that the tests can still be embedded, by caching the existing connection in the SanityContext.

@darkowlzz I liked the API tests that you added to PR #98, so I've copied your commit into this PR and extended the one about embedding so that it actually catches when the tests (incorrectly) try to use their own Before/AfterSuite functions.

### DIFF
--- a/hack/_apitest/api_test.go
+++ b/hack/_apitest/api_test.go
@@ -1,0 +1,18 @@
+package apitest
+
+import (
+	"os"
+	"testing"
+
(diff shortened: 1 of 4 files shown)


lpabon:
Great job both @pohly and @darkowlzz

PR #111, opened 2018-08-31
Add tests for Unicode names

### LINKED ISSUE #108: Need test cases for unicode volume/snapshot names
Although it's not clearly stated, the CSI spec allows Unicode names to be specified for volumes and snapshot. A naively implement plugin might take the suggested name and use it somewhere where Unicode characters aren't supported, and the bug might not get caught if all of the test cases use ASCII names. We should send some non-ASCII names to the plugin and expect success.

bswartz:
Create 2 new tests, one for a volume with a Unicode name, and one
for a snapshot with a Unicode name. These tests are intended to catch
bugs where plugin authors use the value of the name field in a context
where Unicode is not allowed.

Fixes #108

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -443,6 +443,52 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 			Expect(err).NotTo(HaveOccurred())
 			cl.UnregisterVolume(name)
 		})
+
+		It("should not fail when creating volume with a unicode name", func() {
+
+			name := "sanity-unicode-\xF0\x9F\x92\xA9-name"
+			By("creating a volume")
+			size := TestVolumeSize(sc)
+
+			vol, err := c.CreateVolume(
+				context.Background(),
+				&csi.CreateVolumeRequest{
+					Name: name,
+					VolumeCapabilities: []*csi.VolumeCapability{
+						{
+							AccessType: &csi.VolumeCapability_Mount{
+								Mount: &csi.VolumeCapability_MountVolume{},
+							},
+							AccessMode: &csi.VolumeCapability_AccessMode{
+								Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+							},
+						},
+					},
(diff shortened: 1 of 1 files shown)


pohly:
Is a driver allowed to fail when presented with a name that is non-ASCII? I don't think so, because having such restrictions in the driver would be an interoperability issue with spec-compliant COs or worse, break in random cases when users end up choosing such a name.

Therefore I think it is okay to have this test. To me, the csi-test suite should not just check for the bare minimum functionality, but also corner cases like this.

If a driver developer has a legitimate reason to limit the capabilities of his driver such that it can't pass this test, then we can add a configuration option to disable the test so that the suite remains useful for that driver. But I don't think we should worry too much about that yet, i.e. not add such a config option already now.

lpabon:
> Is a driver allowed to fail when presented with a name that is non-ASCII? 
Please do not misunderstand my comment. I mean, if we want this to be a test, it *MUST* be in the spec. _Then_ csi-sanity will test it.

lpabon:
@bswartz Do you mind creating a PR in the CSI spec updating it to mention the support of Unicode? I would like this test to be added to csi-sanity, but I would like to make sure it is a reflection of the spec.

bswartz:
I did raise an issue on the CSI spec.

https://github.com/container-storage-interface/spec/issues/267

The conclusion was that the current wording of the spec allows for *any* UTF-8 string of 128 bytes or less in name fields. I personally think this should be tightened up, but my main issue is that I don't like allowing control characters ('\x00' - '\x1f') in name fields. I have no problem with allowing Unicode names. Either way, it sounds like this issue has already been litigated and they don't intend to change the wording.

pohly:
Ben Swartzlander <notifications@github.com> writes:

> I did raise an issue on the CSI spec.
>
> https://github.com/container-storage-interface/spec/issues/267

Can you add that link to a comment for the new test case in the source
code? It'll help users of the suite understand why they have to support
such names.

lpabon:
I'm closing this issue. Please feel free to open it again if it is part of the CSI spec.

PR #116, opened 2018-09-13
Fix snapshot test

jingxu97:
Remove snapshot invalid token test, modify the workflow for creating a
snapshot with the same name but different volume source id in order to
use valid volume id.

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -1254,21 +1254,6 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *SanityConte
 		Expect(snapshots.GetEntries()).To(BeEmpty())
 	})
 
-	It("should fail when an invalid starting_token is passed", func() {
-		vols, err := c.ListSnapshots(
-			context.Background(),
-			&csi.ListSnapshotsRequest{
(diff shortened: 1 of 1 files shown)


jingxu97:
cc @wackxu @xing-yang

pkg/sanity/controller.go
```diff
@@ -1254,21 +1254,6 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *SanityConte
...
-			},
-		)
-		Expect(err).To(HaveOccurred())
-		Expect(vols).To(BeNil())
-
-		serverError, ok := status.FromError(err)
-		Expect(ok).To(BeTrue())
-		Expect(serverError.Code()).To(Equal(codes.Aborted))
-	})
-
```
xing-yang: I think we still need some kind of invalid token test?  I thought you want to remove the next test "should fail when the starting_token is greater than total number of snapshots".

xing-yang: Just talked to Jing.  We don't know what kind of token a driver will use so we don't really know what is invalid.  So this test can be removed.

jingxu97:
cc @lpabon @saad-ali Could you please help approve? Thanks!

PR #117, opened 2018-09-17
Update README for sanity test using Ginkgo.

leakingtapan:
The original sample is using BeforeEach/AfterEach in a problematic way.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ cmd/csi-sanity/csi-sanity
 
 # JetBrains GoLand
 .idea
+
+# Vim
+*.swp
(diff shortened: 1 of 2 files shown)


pohly:
I still don't like mixing different things in the same commit ;-} Here that's the .gitignore update and the README.md update. Both make sense to me, though, so thanks for the PR.

leakingtapan:
@pohly How about I split into two commits but reuse the same pr?

pohly:
Cheng Pan <notifications@github.com> writes:
> @pohly How about I split into two commits but reuse the same pr?

That's fine with me.

leakingtapan:
@pohly change is made

PR #120, opened 2018-09-26
Updating vendored CSI spec and associated services/mocks

mattsmithdatera:
This change is needed to keep CSI drivers that are working off the latest CSI spec from failing the sanity tests.  For example a CSI driver that implements GetVolumeStats will fail the Sanity test for GetCapabilities since it has no knowledge of the new GetVolumeStats capability.


mattsmithdatera:
This needs to be merged into master, not 0.3.0.  Closing and creating a new PR

Issue #127, opened 2018-10-23
Images on quay.io can't run


jsafrane:
```
$ docker run -ti quay.io/k8scsi/mock-driver:canary
standard_init_linux.go:190: exec user process caused "no such file or directory"
```

The reason is that `mock` binary inside the container can't be dynamically linked:

```
$  docker run -ti --entrypoint /bin/sh quay.io/k8scsi/mock-driver:canary
(inside the container)#  ldd mock
        /lib64/ld-linux-x86-64.so.2 (0x7fa22001d000)
        libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7fa22001d000)
        libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7fa22001d000)

(inside the container)# ls /lib64/ld-linux-x86-64.so.2 
ls: /lib64/ld-linux-x86-64.so.2: No such file or directory
```

I've no clue how the image was built, perhaps some hiccup on travis.

jsafrane:
/reopen

It's still broken in v0.3.0-2:

```
$ docker run -ti quay.io/k8scsi/mock-driver:v0.3.0-2
standard_init_linux.go:178: exec user process caused "no such file or directory"
```

today's canary is the same.

PR #128, opened 2018-10-23
Fix canary builds

jsafrane:
This PR makes travis.yaml look like in the other kubernetes-csi repos. As side-effect, new image should be built to fix https://github.com/kubernetes-csi/csi-test/issues/127

Fixes: https://github.com/kubernetes-csi/csi-test/issues/127

@pohly @lpabon

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,16 @@
 language: go
+sudo: required
+services:
+  - docker
 matrix:
   include:
   - go: 1.10.3
 script:
 - make test
+- make
 after_success:
   - if [ "${TRAVIS_BRANCH}" == "master" ] && [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then
+      make container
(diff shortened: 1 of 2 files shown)


pohly:
/lgtm

But let me check with @lpabon first before I approve - not because of anything in this PR, just because we haven't discussed which PRs he wants me to approve and which ones he wants to be involved in.

.travis.yml
```diff
@@ -1,11 +1,16 @@
 language: go
+sudo: required
+services:
+  - docker
 matrix:
   include:
-  - go: 1.10.3
+  - go: 1.x
```
davidz627: would recommend pinning to a specific version. I believe we pinned to `1.10.3` before because `1.11` was breaking everything at the time and we could not submit any other PR's before resolving that issue

jsafrane: I'd swear I took 1.x from another kubernetes-csi repo, but I can't find it now. Restored 1.10.3.

pohly:
Looks like @davidz627 comment about the Go version was addressed, so:
/lgtm
and
/approve

PR #129, opened 2018-10-25
Add more description to mock CSI driver

jsafrane:
Primary goal is to test travis, but when I need to create some useful PR :-)


lpabon:
Looks good to me

/lgtm
/approve

PR #130, opened 2018-10-25
Use binary complied by make for e2e tests.

jsafrane:
`hack/e2e.sh` should use the "official" binary created by make instead of its own build

1. we should test the real binary
2. `make test; make; make container` (as our travis does) should use the real binary instead of the wrong one built by `make test`

@pohly @lpabon, PTAL

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ container: $(APP)
 push: container
 	docker push $(IMAGE_NAME):$(IMAGE_VERSION)
 
-test:
+test: $(APP)
 	files=$$(find ./ -name '*.go' | grep -v '^./vendor' ); \
         if [ $$(gofmt -d $$files | wc -l) -ne 0 ]; then \
                 echo "formatting errors:"; \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -38,7 +38,7 @@ container: $(APP)
 push: container
 	docker push $(IMAGE_NAME):$(IMAGE_VERSION)
 
-test:
+test: $(APP)
```
pohly: The commit message says "make test; make; make container", but that won't work anymore, right? It's perhaps better to put "make; make test; make container" into the commit message and mention in the Makefile for "test" that one has to do "make" first.

jsafrane: I think the order does not really matter. All three `make`, `make test` and `make container` depend on `$(APP)`. So the first one executed compiles the binary, the others re-use it. You may notice I removed `make` in this PR, because it's useless.

pohly: Right, it works here the way you said.

I was expecting "$(APP)" to be a phony target because there is no real dependency tracking for Go+make, i.e. "make $(APP)" always has to invoke "go build" to ensure that the binary really is up-to-date. But this Makefile here only builds the binary if it doesn't exist yet, without rebuilding it when it is out-dated.

I guess that's how it is meant to be here, so /lgtm.

Issue #132, opened 2018-11-01
Need to delete snapshot before deleting volume


wnxn:
Hi,
It is better to delete snapshot before deleting volume at cleaning up stage of ListSnapshots testcase. So, we could change the cleaner: 
https://github.com/kubernetes-csi/csi-test/blob/3e3231b4e4af4cf62615a435ca4dd0e70d63fbde/pkg/sanity/controller.go#L1208-L1216

to：
https://github.com/kubernetes-csi/csi-test/blob/3e3231b4e4af4cf62615a435ca4dd0e70d63fbde/pkg/sanity/controller.go#L1251-L1259

PR #133, opened 2018-11-01
Fix listsnapshot clean up

wnxn:
#132

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -1205,14 +1205,14 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *SanityConte
 		verifySnapshotInfo(snapshots.GetEntries()[0].GetSnapshot())
 		Expect(snapshots.GetEntries()[0].GetSnapshot().GetId()).To(Equal(snapshot.GetSnapshot().GetId()))
 
-		By("cleaning up deleting the volume")
-		delVolReq := MakeDeleteVolumeReq(sc, volume.GetVolume().GetId())
-		_, err = c.DeleteVolume(context.Background(), delVolReq)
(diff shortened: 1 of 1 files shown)


wnxn:
Hi, Patrick
In some storage server, snapshot is related on a specific volume. If user delete the volume, the snapshot will 
 be deleted automatically at storage server side.  For this reason, we must keep the order of deletion, that is deleting volume‘s snapshots before deleting the volume. It's found that the author might take the order of deletion into consideration when writing other snapshot test cases.
Just like this:
https://github.com/kubernetes-csi/csi-test/blob/3e3231b4e4af4cf62615a435ca4dd0e70d63fbde/pkg/sanity/controller.go#L1251-L1259

pohly:
But what breaks when the order is as it is now? Does the test fail?

I am trying to understand whether this change works around unexpected behavior of one particular storage backend or is actually broken.

@xing-yang can you perhaps comment on this?

xing-yang:
For some storage systems, snapshot has to be deleted first before deleting the volume.  If you delete a volume with dependent snapshot, it may cause error in some storage systems.  So this change looks good to me.

pohly:
@wnxn I'm fine with merging this then, but I'd still like to have this information recorded in the commit message. Please use "git commit --amend" and "git push ... +master" to update the PR.

wnxn:
Thanks for @pohly and @xing-yang 's review. 
I add some clarification in commit message. Would you please review again?

pohly:
/lgtm

Just one comment for future commits: please format commit messages with explicit line breaks in the body, they are easier to read in "git log" then.

wnxn:
Hi, @pohly 
Thanks for your advice. I will format messages body in my future commits.

PR #135, opened 2018-11-08
Add driver roles to sanity tests

jarrpa:
This patch adds the ability to specify driver roles (Identity,
Controller, Node) for a given test run. This allows for a convenient,
code-based mechanism to parse which sanity tests will be run for a given
driver. This then allows for simpler "go test" CLI calls that helps
facilitate development and automation complexity.

Signed-off-by: Jose A. Rivera <jarrpa@redhat.com>

### DIFF
--- a/pkg/sanity/sanity.go
+++ b/pkg/sanity/sanity.go
@@ -32,6 +32,24 @@ import (
 	. "github.com/onsi/gomega"
 )
 
+// Valid CSI driver roles, used for parsing tests
+type DriverRole string
+type DriverRoles []DriverRole
+
+func (drs DriverRoles) String() []string {
+	strs := []string{}
+	for _, d := range drs {
+		strs = append(strs, string(d))
+	}
+	return strs
+}
+
+const (
+	DriverRoleIdentity   DriverRole = "Identity"
+	DriverRoleController DriverRole = "Controller"
+	DriverRoleNode       DriverRole = "Node"
+)
+
 // CSISecrets consists of secrets used in CSI credentials.
 type CSISecrets struct {
 	CreateVolumeSecret              map[string]string `yaml:"CreateVolumeSecret"`
@@ -51,6 +69,8 @@ type Config struct {
 	StagingPath string
 	Address     string
 	SecretsFile string
+
+	Roles DriverRoles
 
 	TestVolumeSize           int64
 	TestVolumeParametersFile string
--- a/pkg/sanity/tests.go
+++ b/pkg/sanity/tests.go
@@ -17,6 +17,10 @@ limitations under the License.
 package sanity
 
 import (
+	"fmt"
+	"regexp"
+	"strings"
+
 	. "github.com/onsi/ginkgo"
 )
 
@@ -40,17 +44,20 @@ func DescribeSanity(text string, body func(*SanityContext)) bool {
 // registerTestsInGinkgo invokes the actual Gingko Describe
 // for the tests registered earlier with DescribeSanity.
 func registerTestsInGinkgo(sc *SanityContext) {
+	roleMatch := regexp.MustCompile(fmt.Sprintf(".*(?:%s).*", strings.Join(sc.Config.Roles.String(), "|")))
 	for _, test := range tests {
-		Describe(test.text, func() {
-			BeforeEach(func() {
-				sc.setup()
-			})
+		if roleMatch.MatchString(test.text) {
+			Describe(test.text, func() {
+				BeforeEach(func() {
+					sc.setup()
+				})
 
-			test.body(sc)
+				test.body(sc)
 
-			AfterEach(func() {
-				sc.teardown()
+				AfterEach(func() {
+					sc.teardown()
+				})
 			})
-		})
+		}
 	}
 }


jarrpa:
@lpabon PTAL

pkg/sanity/tests.go
```diff
@@ -40,17 +44,20 @@ func DescribeSanity(text string, body func(*SanityContext)) bool {
 // registerTestsInGinkgo invokes the actual Gingko Describe
 // for the tests registered earlier with DescribeSanity.
 func registerTestsInGinkgo(sc *SanityContext) {
+	roleMatch := regexp.MustCompile(fmt.Sprintf(".*(?:%s).*", strings.Join(sc.Config.Roles.String(), "|")))
 	for _, test := range tests {
-		Describe(test.text, func() {
-			BeforeEach(func() {
-				sc.setup()
-			})
+		if roleMatch.MatchString(test.text) {
```
pohly: Would it make sense to use
```
if !roleMatch.MatchString(test.text) {
    continue
}
````
to avoid indenting the rest of the code?

Just thinking aloud here, I am not sure whether it is considered better Go style.

File: pkg/sanity/sanity.go [hunk 2]
```diff
@@ -32,6 +32,24 @@ import (
...
+	for _, d := range drs {
+		strs = append(strs, string(d))
+	}
+	return strs
+}
+
+const (
+	DriverRoleIdentity   DriverRole = "Identity"
+	DriverRoleController DriverRole = "Controller"
+	DriverRoleNode       DriverRole = "Node"
```
pohly: It is not obvious that these strings are sub-strings of the first parameter of each `DescribeSanity` call. Perhaps this should be made explicit and enforced by turning the first parameter of that function into a `DriverRole` and then appending the common ` Service` suffix in the implementation of that call.

pkg/sanity/tests.go
jarrpa: No idea about "better" Go style. I avoided it for my own personal aesthetics, since I like to avoid explicit flow control statements where possible. :)

pkg/sanity/sanity.go
jarrpa: I was explicitly trying to avoid this, because I didn't want to add a new parameter to the test functions nor go and update every single existing testing function to a new format. I can add a comment above them to describe this relationship, if that would help?

pohly: Such a comment would have to be added to `sanity.go` and all calls to `DescribeSanity`, because otherwise someone looking at those calls won't know that the strings have a special meaning. Then one might as well change the call.

pkg/sanity/tests.go
pohly: I've been asked a few times to rewrite code such that it indents less, so having explicit "return" or "continue" statements might indeed be the preferred style.

Anyway, it makes the diff smaller, so please change it as suggested above unless it hurts your aesthetics too much ;-}

lpabon:
@jarrpa I'm not sure what you mean by this test. What is a driver role and what is the use case you are trying to satisfy?

jarrpa:
What I call a driver role is one of Node, Controller, or Identity. This is to satisfy a use case in which a driver is comprised of multiple binaries that do not all implement all three roles. But, this use case may be being thrown out the window, so there may not be a need for this anymore. I'll update or close this PR as that changes.

pohly:
"Jose A. Rivera" <notifications@github.com> writes:
> What I call a driver role is one of Node, Controller, or
> Identity. This is to satisfy a use case in which a driver is comprised
> of multiple binaries that do not all implement all three roles. But,
> this use case may be being thrown out the window, so there may not be
> a need for this anymore. I'll update or close this PR as that changes.

A better solution for that use case would be to support a second address
for controller which then (if set) gets used for connections to the
controller. That way, tests that need both (like provisioning a volume
and then staging it) can still run when the driver is split.

Regarding identity, isn't that something that always has to be
implemented in the same binary and on the same connection as controller
and node?

jarrpa:
> A better solution for that use case would be to support a second address for controller which then (if set) gets used for connections to the controller. That way, tests that need both (like provisioning a volume and then staging it) can still run when the driver is split.

What about for the case where there is a controller with no node server?

> Regarding identity, isn't that something that always has to be implemented in the same binary and on the same connection as controller and node?

Sure, I just split it out initially because it is logically separate at the same layer.

pohly:
"Jose A. Rivera" <notifications@github.com> writes:

>> A better solution for that use case would be to support a second address for controller which then (if set) gets used for connections to the controller. That way, tests that need both (like provisioning a volume and then staging it) can still run when the driver is split.
>
> What about for the case where there is a controller with no node
> server?

We could use:

// Controller and node under test, combined implementation (the traditional mode):
Config {
        Address: "/tmp/my-combined-node-controller.sock",
}

// Controller and node under test, split implementation.
Config {
       Address: "/tmp/my-node.sock",
       AddressController: "/tmp/my-controller.sock",
}

// Only controller under test.
Config {
       Address: ""
       AddressController: "/tmp/my-controller.sock",
}

That's the three different configurations that IMHO are worth
supporting.

jarrpa:
So something like:
```
Config {
  Address           string # controller+node driver
  AddressController string # controller-only driver
  AddressNode       string # node-only driver
}
```
Possibly also with a check that if `Address` is defined the other two must be unset?

pkg/sanity/tests.go
davidz627: idiomatic go code (as well as my preferred style) is to have explicit returns/continues to avoid having large blocks of code indented. It makes it very clear under "what condition" the function should terminate early and not continue doing any execution whatsoever in that function. With large blocks of indented code you run the risk of having the scroll down before noticing that there may be more code after the closing brace of the indented code which hurts readability.

jarrpa: But such code tends to lead to higher cyclomatic complexity, which then leads to a tendency to have microfunctions as a way to combat that, which further hurts readability.

jarrpa:
See the words "tends to". :P I just know that I've been able to greatly reduce cyclomatic complexity in various places by refactoring to remove unnecessary conditionals. Regardless, I'm all for adopting community conventions, I'll just keep thinking they're misguided when taken as absolutes. ;)

pkg/sanity/tests.go
davidz627: I don't believe this changes the cyclomatic complexity. It merely inverts the cases, does not add or take away any

jarrpa: I never claimed it did.

pohly:
@jarrpa the support for specifying controller and node address separately has been merged (https://github.com/kubernetes-csi/csi-test/pull/158).

Do you still think that we need this PR?

jarrpa:
Nope!

PR #136, opened 2018-11-13
Use NodeGetInfo to get node Id

wnxn:
Because NodeGetId RPC call is deprecated in CSI spec v0.3.0,
we should use NodeGetInfo to get node Id in sanity test.

In the part of [NodeGetId in CSI spec v0.3.0](https://github.com/container-storage-interface/spec/blob/v0.3.0/spec.md#nodegetid), it is found that NodeGetId has been deprecated.
```
NodeGetId RPC call is deprecated. Users of this RPC call SHOULD use NodeGetInfo.
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -787,10 +787,10 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 			Expect(vol.GetVolume().GetId()).NotTo(BeEmpty())
 			cl.RegisterVolume(name, VolumeInfo{VolumeID: vol.GetVolume().GetId()})
 
-			By("getting a node id")
-			nid, err := n.NodeGetId(
+			By("getting a node info")
+			nid, err := n.NodeGetInfo(
 				context.Background(),
-				&csi.NodeGetIdRequest{})
+				&csi.NodeGetInfoRequest{})
 			Expect(err).NotTo(HaveOccurred())
 			Expect(nid).NotTo(BeNil())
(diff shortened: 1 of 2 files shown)


lpabon:
@wnxn @pohly I'm not sure if this necessary. In the master branch (v1.1.0) and in the release-1.0 branch, we now use NodeGetInfo, and only in the release-0.3 branch do we use NodeGetId.

wnxn:
Hi @pohly and @lpabon ,
For CSI test release-1.0, I notice that we only call NodeGetInfo now.
For CSI test release-0.3, if this project still support to test CSI plugins based on CSI spec v0.3.0. I believe we should call NodeGetInfo to get node Id when setting `ControllerPublishVolumeRequest`'s `node_id` field. 
In [CSI spec v0.3.0](https://github.com/container-storage-interface/spec/blob/v0.3.0/csi.proto#L572),  developer would be better to follow this advice, the CO SHALL set `ControllerPublishVolumeRequest`'s `node_id` field to match the node ID returned by `NodeGetInfo`. At least, we should get node Id through NodeGetInfo instead of NodeGetId when setting `ControllerPublishVolumeRequest`'s `node_id` field.

```
message ControllerPublishVolumeRequest {
  ...
  // The ID of the node. This field is REQUIRED. The CO SHALL set this
  // field to match the node ID returned by `NodeGetInfo`.
  string node_id = 2;
  ...
}
```

pohly:
We are using `NodeGetInfo` on master and rely on CSI 1.0. I think we can close this now.

/close

Issue #137, opened 2018-11-13
Broken release `v0.3.0-3`


fatih:
Hi,

I was using `v0.3.0-2` for https://github.com/digitalocean/csi-digitalocean with the CSI Spec `v0.3.0` tagged release (see: https://github.com/container-storage-interface/spec/releases/tag/v0.3.0).

I decided to upgrade to `v0.3.0-3` as it was including the latest changes.  However when I upgrade it, my test are failing due the following error:

```
# github.com/kubernetes-csi/csi-test/pkg/sanity
../../go/pkg/mod/github.com/kubernetes-csi/csi-test@v0.3.0-3/pkg/sanity/node.go:122:10: undefined: csi.NodeServiceCapability_RPC_GET_VOLUME_STATS
```

When I check the CSI `v0.3.0` version, there is no such identifier. This was added later to CSI master and is part of the currently released `v1.0.0-rc1` release (https://github.com/container-storage-interface/spec/releases/tag/v1.0.0-rc1). 

When I update my CSI spec to `v1.0.0-rc1` from `v0.3.0`, it still fails, albeit with a different error:

```
go: import "github.com/digitalocean/csi-digitalocean/driver" ->
        test ->
        import "github.com/kubernetes-csi/csi-test/pkg/sanity" ->
        import "github.com/container-storage-interface/spec/lib/go/csi/v0": looping trying to add package
```

The reason is, CSI `v1.0.0-rc1` get rid of the `/v0` suffix from their import path, but the CSI-sanity tests still uses the `/v0` suffix, example: https://github.com/kubernetes-csi/csi-test/blob/master/driver/driver.go#L32

Because of this, the `v0.3.0-3` is not compatible with either `v0.3.0` or `v1.0.0-rc1` of the CSI spec.

I think a new `v0.3.0-4` version should be released that removes the `csi.NodeServiceCapability_RPC_GET_VOLUME_STATS` identifier because it's not part of `v0.3.0`, or we should release a csi-test `v1.0.0-rc1` release that is compatible with CSI spec `v1.0.0-rc1` release that was released today.

lpabon:
Thank you for bringing this up @fatih . @pohly figured out that the CSI spec had been changed in our repo.  Also, the importing issue, is new to me. I'll have to investigate.

lpabon:
We have released 1.0.0 and are working on a new 0.3.0-4 (#140) release

pohly:
0.3.0-4 has been tagged.

[...]

PR #138, opened 2018-11-14
Update to v1.0.0-RC2 of CSI

lpabon:
Signed-off-by: Luis Pabón <lpabon@gmail.com>


lpabon:
need to update old examples also.. Working on that now..

davidz627:
Sanity and Vendor changes LGTM

msau42:
lgtm, could you separate vendor and non-vendor into different commits?

pohly:
With this PR merged and v1.0.0-rc2 released, should https://github.com/kubernetes-csi/csi-test/issues/137 get closed as "won't fix" because everyone has to move to CSI 1.0.0 now anyway?

What CSI driver developers who still need to support older Kubernetes releases? Does a CSI 1.0 driver work on, say, Kubernetes 1.11?

msau42:
CSI in Kubernetes 1.13 is not compatible with 1.12.  Drivers need to update to the 1.0  spec to work on 1.13.  We should still support older releases of Kubernetes that use the older CSI spec.

PR #141, opened 2018-11-22
Fix controller publish in readonly mode

wnxn:
Plugin's PUBLISH_READONLY capability should be verified
before publishing volume in readonly mode.


wnxn:
/assign @pohly 
/assign @lpabon 
Would you please review this PR ?

PR #145, opened 2018-11-26
Add ControllerServiceCapability_RPC_CLONE_VOLUME to GetCapabilities.

humblec:
Signed-off-by: Humble Chirammal <hchiramm@redhat.com>


humblec:
@lpabon @saad-ali can you please review this PR ?

humblec:
@pohly one other capability which I see from 1.0.0 release is `csi.ControllerServiceCapability_RPC_PUBLISH_READONLY` which is already available in our sanity package.  So, I dont think something else is missing, that way this is good to go. If at all we missed anything we can bring that soon with another PR. I will be running some more iterations anyway to cross check.

PR #150, opened 2019-01-05
sanity: close connection

darkowlzz:
Close the gRPC connection properly at the end of the test to avoid any
transport error in the subsequent connections.

### DIFF
--- a/pkg/sanity/sanity.go
+++ b/pkg/sanity/sanity.go
@@ -89,6 +89,7 @@ func Test(t *testing.T, reqConfig *Config) {
 	registerTestsInGinkgo(sc)
 	RegisterFailHandler(Fail)
 	RunSpecs(t, "CSI Driver Test Suite")
+	sc.Conn.Close()
 }
 
 func GinkgoTest(reqConfig *Config) {


pohly:
The change looks okay, but I'm curious what kind of problem this fixes. https://github.com/kubernetes-csi/csi-test/blob/master/cmd/csi-sanity/sanity_test.go exits directly after calling `Test`, so nothing should run after it which could get confused by the open connection.

Can you explain where the missing `Close` is causing issues?

darkowlzz:
> Can you explain where the missing Close is causing issues?

Hi, I came across this issue when running csi-sanity tests followed by some other custom tests that connect to the same CSI server. When the tests following csi-sanity runs, they failed with error:
```
grpc transport: http2Server.HandleStreams failed to read frame
```
It seemed to happen consistently in the private CI servers, but I couldn't reproduce it on my development machine. After searching for the error on the web, it seemed this happens when the old connections are closed prematurely. And looking at the csi-sanity code, I found that there was no call to properly close the connection.
To solve the issue for now, using the existing release binary, I reordered the tests by running custom tests before csi-sanity and that worked without any grpc error.
I thought it's just better to close the connection properly, hence added a close at the end.

Hope that helps.

pohly:
Makes sense. Something similar might be needed also when running under Ginkgo, but it's non-trivial to add (Ginkgo's `AfterEach` runs more than once and can't be used to clean up after all all tests are done), so let's wait until it actually occurs.

/lgtm
/approve

PR #155, opened 2019-01-08
sanity: Add max volumes attached test

darkowlzz:
This adds a test for ControllerPublishVolume to check the maximum volume
attachment on a node.
It also sets max attach limit in the mock driver to 2 and adds a check
in the driver for the number of volume attachments before publishing a
new volume.

ControllerPublishVolume [Max volumes attached](https://github.com/container-storage-interface/spec/blob/release-1.0/spec.md#controllerpublishvolume-errors) error.

### DIFF
--- a/mock/service/controller.go
+++ b/mock/service/controller.go
@@ -166,6 +166,11 @@ func (s *service) ControllerPublishVolume(
 		}, nil
 	}
 
+	// Check attach limit before publishing only if attach limit is set.
+	if s.config.AttachLimit > 0 && s.getAttachCount(devPathKey) >= s.config.AttachLimit {
+		return nil, status.Errorf(codes.ResourceExhausted, "Cannot attach any more volumes to this node")
+	}
+
 	var roVal string
 	if req.GetReadonly() {
 		roVal = "true"
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -895,6 +895,48 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 			cl.UnregisterVolume(name)
 		})
 
+		It("should fail when publishing more volumes than the node max attach limit", func() {
+			if !sc.Config.TestNodeVolumeAttachLimit {
+				Skip("testnodevolumeattachlimit not enabled")
+			}
+
+			By("getting node info")
+			nodeInfo, err := n.NodeGetInfo(
+				context.Background(),
+				&csi.NodeGetInfoRequest{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(nodeInfo).NotTo(BeNil())
+
+			if nodeInfo.MaxVolumesPerNode <= 0 {
+				Skip("No MaxVolumesPerNode")
+			}
+
+			nid := nodeInfo.GetNodeId()
+			Expect(nid).NotTo(BeEmpty())
+
+			// Store the volume name and volume ID for later cleanup.
+			createdVols := map[string]string{}
+			By("creating volumes")
+			for i := int64(0); i < nodeInfo.MaxVolumesPerNode; i++ {
+				name := uniqueString(fmt.Sprintf("sanity-max-attach-limit-vol-%d", i))
+				volID, err := CreateAndControllerPublishVolume(sc, c, name, nid)
+				Expect(err).NotTo(HaveOccurred())
+				cl.RegisterVolume(name, VolumeInfo{VolumeID: volID, NodeID: nid})
+				createdVols[name] = volID
+			}
+
+			extraVolName := uniqueString("sanity-max-attach-limit-vol+1")
+			_, err = CreateAndControllerPublishVolume(sc, c, extraVolName, nid)
+			Expect(err).To(HaveOccurred())
+
+			By("cleaning up")
+			for volName, volID := range createdVols {
+				err = ControllerUnpublishAndDeleteVolume(sc, c, volID, nid)
+				Expect(err).NotTo(HaveOccurred())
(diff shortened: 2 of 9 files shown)


pkg/sanity/controller.go
```diff
@@ -854,6 +854,43 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
...
+			cl.RegisterVolume(name1, VolumeInfo{VolumeID: volID1, NodeID: nid})
+
+			name2 := uniqueString("sanity-max-attach-limit-vol2")
+			volID2, err := CreateAndControllerPublishVolume(sc, c, name2, nid)
+			Expect(err).NotTo(HaveOccurred())
+			cl.RegisterVolume(name2, VolumeInfo{VolumeID: volID2, NodeID: nid})
+
+			name3 := uniqueString("sanity-max-attach-limit-vol3")
+			_, err = CreateAndControllerPublishVolume(sc, c, name3, nid)
+			Expect(err).To(HaveOccurred())
```
pohly: This seems to assume that the driver has an attach limit of 2. This isn't going to be true for all drivers that are getting sanity tested.

The test needs to check the limit by querying the driver, skip the test if there is no limit, otherwise create n + 1 volumes to reach the limit.

darkowlzz: Right! I was thinking in terms of the mock driver only :sweat_smile:
Thanks a lot for pointing that out.

darkowlzz: Updated! Should be good now. :slightly_smiling_face:

pohly: The world would be a simpler place if we only wrote software where test code only talks to other test code. But alas, there's also such a thing as production code :slightly_smiling_face:

Thanks for addressing this.

mock/service/controller.go
```diff
@@ -166,6 +166,11 @@ func (s *service) ControllerPublishVolume(
 		}, nil
 	}
 
+	// Check attach limit before publishing only if attach limit is set.
+	if int(s.config.AttachLimit) > 0 && s.getAttachCount(devPathKey) >= s.config.AttachLimit {
+		return nil, status.Errorf(codes.ResourceExhausted, "Cannot attach any more volumes to this node")
```
pohly: Why are you casting `s.config.AttachLimit` here?

If I remember correctly, `0` is an untyped integer in Go, it should compare fine to both int64 and normal int.

pkg/sanity/controller.go
```diff
@@ -854,6 +855,45 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 			cl.UnregisterVolume(name)
 		})
 
+		It("should fail when publishing volumes more than the node max attach limit", func() {
```
pohly: "should fail when publishing more volumes than allowed by the node max attach limit"

pkg/sanity/controller.go
```diff
@@ -854,6 +855,45 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
...
+			By("getting node info")
+			nodeInfo, err := n.NodeGetInfo(
+				context.Background(),
+				&csi.NodeGetInfoRequest{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(nodeInfo).NotTo(BeNil())
+
+			if int(nodeInfo.MaxVolumesPerNode) <= 0 {
+				Skip("No MaxVolumesPerNode")
+			}
```
pohly: Same here: type cast really necessary?

pkg/sanity/controller.go
```diff
@@ -854,6 +855,45 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
...
+
+			nid := nodeInfo.GetNodeId()
+			Expect(nid).NotTo(BeEmpty())
+
+			// Store the volume name and volume ID for later cleanup.
+			createdVols := map[string]string{}
+			By("creating volumes")
+			for i := 0; i < int(nodeInfo.MaxVolumesPerNode); i++ {
+				name := uniqueString(strings.Join([]string{"sanity-max-attach-limit-vol", strconv.Itoa(i)}, "-"))
```
pohly: I think `fmt.Sprintf` will be easier to read than this `strings.Join` + `strconv.Itoa`.

pkg/sanity/controller.go
```diff
@@ -854,6 +855,45 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
...
+				Skip("No MaxVolumesPerNode")
+			}
+
+			nid := nodeInfo.GetNodeId()
+			Expect(nid).NotTo(BeEmpty())
+
+			// Store the volume name and volume ID for later cleanup.
+			createdVols := map[string]string{}
+			By("creating volumes")
+			for i := 0; i < int(nodeInfo.MaxVolumesPerNode); i++ {
```
pohly: This is problematic: consider what happens when nodeInfo.MaxVolumesPerNode is an int64 which is larger than the maximum `int`. Better ensure that `i` is `int64`, with:
```
for i := int64(0); i < nodeInfo.MaxVolumesPerNode; i++ {
```

mock/service/controller.go
darkowlzz: Thanks! I wasn't aware of that. Thought casting would be better.
Updated now.

darkowlzz:
Thanks! I've squashed all the commits.

pohly:
@darkowlzz sorry, should have merged this sooner, now you need to rebase. Can you do that and then I'll merge?

darkowlzz:
@pohly np :slightly_smiling_face: . Rebased.

PR #158, opened 2019-01-26
Add support for separate controller and node addresses

### LINKED ISSUE #142: sanity: support different addresses for controller and node operations
Currently the sanity pkg connects to the CSI driver at a single address and expects to find all services there. Not all CSI drivers are deployed like that and some might not even support all services in the same binary. Therefore I propose to have Config.Address (as it is now) and to add an optional Config.ControllerAddress that gets used if set when connecting to the controller.

darkowlzz:
This change adds a new flag `controllerendpoint` to csi-sanity which
could be used to pass a separate controller endpoint for CSI drivers
that have separate controller and node services

It also modifies the mock driver to support separate controller and node
component mode when CSI_CONTROLLER_ENDPOINT env var is passed to it.
Two separate unix domain sockets are created and the csi-sanity connects
to the appropriate socket while running the RPCs.

Adds a new e2e test section to test this setup.

Fixes #142

### DIFF
--- a/cmd/csi-sanity/sanity_test.go
+++ b/cmd/csi-sanity/sanity_test.go
@@ -36,6 +36,7 @@ var (
 
 func init() {
 	flag.StringVar(&config.Address, prefix+"endpoint", "", "CSI endpoint")
+	flag.StringVar(&config.ControllerAddress, prefix+"controllerendpoint", "", "CSI controller endpoint")
 	flag.BoolVar(&version, prefix+"version", false, "Version of this program")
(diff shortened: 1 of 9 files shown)


shay-berman:
Hi Please see my comment -> https://github.com/kubernetes-csi/csi-test/issues/142#issuecomment-469699437
maybe it will be relevant to add it as part of this PR.  Let me know what do you think

pohly:
@darkowlzz sorry, please rebase.

And please, don't hesitate to ping me if a PR was reviewed and then doesn't get merged.

darkowlzz:
Rebased.

darkowlzz:
The mock driver binary name changed and that's resulting in the failing. Updating the name.

PR #159, opened 2019-01-27
Make mount and staging dir creation flexible

### LINKED ISSUE #144: sanity: more flexible mount and staging directory creation
When testing against a driver that was deployed in a real cluster, creating the mount and staging directories has to happen inside the cluster on the node on which the driver under testing is running. Doing that is specific to the test, so the sanity package should leave the work to some callbacks in sanity.Config if they are defined (must remain optional).

darkowlzz:
This change adds an option to attach callback functions to CSI sanity
for mount target directory creation. This could be used to customize the
target and staging mount path creation.

Also, adds `targetPath` and `stagingPath` fields to `SanityContext` to
store the actual target and staging paths, derived from the sanity
config.

Adds an e2e test section to test the above setup.

Fixes #144


pkg/sanity/sanity.go
```diff
@@ -55,6 +55,13 @@ type Config struct {
 	TestVolumeSize           int64
 	TestVolumeParametersFile string
 	TestVolumeParameters     map[string]string
+
+	// Callback functions to customize the creation of target and staging
+	// directories. Returns the new paths for mount and staging.
+	// If not defined, directories are created in the default way at TargetPath
+	// and StagingPath.
+	CreateTargetDir  func(path string) (string, error)
+	CreateStagingDir func(path string) (string, error)
```
pohly: I'm fine with not exposing this functionality in the command line tool.

If someone needs it, we could add it by letting the test command invoke an external tool.

pkg/sanity/sanity.go
```diff
@@ -123,11 +134,28 @@ func (sc *SanityContext) setup() {
...
 		Expect(err).NotTo(HaveOccurred())
+		sc.targetPath = sc.Config.TargetPath
+	}
+
+	if len(sc.Config.StagingPath) > 0 {
+		// If callback function for creating staging dir is specified, use it.
+		if sc.Config.CreateStagingDir != nil {
+			newpath, err := sc.Config.CreateStagingDir(sc.Config.StagingPath)
+			Expect(err).NotTo(HaveOccurred())
+			sc.stagingPath = newpath
```
pohly: Always explain what the error is about, see https://github.com/kubernetes/kubernetes/issues/34059. I know, the rest of the code doesn't do that either (would be a good enhancement), but it's never too late to do it better.

This could is rather repetitive. I would change the `createMountTargetLocation` method and then use it like this:

```
sc.targetPath = createMountTargetLocation(sc.Config.TargetPath, sc.Config.CreateTargetDir)
sc.stagingPath = createMountTargetLocation(sc.Config.StagingPath, sc.Config.CreateStagingDir)
```

Speaking of `createMountTargetLocation`: there is a call to it in `node.go` which causes directories to be created on the host even when `sc.Config.StagingPath` is set. I think that call can be removed.

okartau: I can confirm that in scenario we use (staging and target directories are created on VMs where pods run, host acting only as test orchestrator) I see /var/tmp/tmp.XXXXXXXXXX directories created on host which are not used and remain undeleted, both with csi-test 1.0.2 and with current PR change.
Positive feedback is that tests pass same way with current PR changes (as with 1.0.2)

darkowlzz:
Updated the PR addressing the suggestions.

Changed `runTestAPIWithCustomTargetPaths()` to run `NodePublishVolume` based tests only to have a deterministic number of target and staging directory creation calls.

Added more context to the test failure errors and changed the signature of `createMountTargetLocation()` to repeat repetition.

pkg/sanity/sanity.go
```diff
@@ -156,18 +170,35 @@ func (sc *SanityContext) teardown() {
...
-func createMountTargetLocation(targetPath string) error {
-	fileInfo, err := os.Stat(targetPath)
-	if err != nil && os.IsNotExist(err) {
-		return os.MkdirAll(targetPath, 0755)
-	} else if err != nil {
-		return err
+func createMountTargetLocation(targetPath string, customCreateDir func(string) (string, error)) (string, error) {
+
+	// Return the target path if empty.
+	if len(targetPath) < 0 {
```
pohly: An empty string has length 0, so this check is not getting triggered as intended.

Better compare with `if targetPath == ""`. It's more obvious what's meant (always most important) and there's no obvious performance benefit from the length comparison (performance doesn't matter here anyway).

pkg/sanity/sanity.go
```diff
@@ -156,18 +170,35 @@ func (sc *SanityContext) teardown() {
...
+	if customCreateDir != nil {
+		newpath, err := customCreateDir(targetPath)
+		if err != nil {
+			return "", err
+		}
+		newTargetPath = newpath
+	} else {
+		fileInfo, err := os.Stat(targetPath)
+		if err != nil && os.IsNotExist(err) {
+			return "", os.MkdirAll(targetPath, 0755)
```
pohly: Here you return an empty string also when `os.MkdirAll` succeeds. That can't be right? I wonder why it isn't breaking the Travis CI check. Ah, probably because the mock driver doesn't actually use the directories and doesn't check whether they exist (unverified). Might be worth adding such a check. Perhaps add such a check as part of the PR, verify "make test" fails, then fix this code?

Overall I find this version more readable:
```
		fileInfo, err := os.Stat(targetPath)
		if err != nil {
                     if !os.IsNotExist(err) {
                        return "", err
                     }
	             if err := os.MkdirAll(targetPath, 0755); err != nil {
                          return "", err
                     }
                }
		if !fileInfo.IsDir() {
			return "", fmt.Errorf("Target location %s is not a directory", targetPath)
		}

**Comment by pohly** [comment]
I said earlier that adding support for this to the command line can wait until we have a need for it. I now found a use case, so can you add command line flags for this? They should accept a single parameter, the command to run, and then call that command with the path as argument. stdout of the command is the new  path, stderr can go to stderr of the caller (i.e. don't redirect).

My use case is for adding sanity testing to https://github.com/kubernetes-csi/csi-driver-host-path/pull/16: when building the hostpath driver (or soon, also other drivers), I want to check out  a certain revision of csi-test and run the sanity command against the deployed driver, without having to built a custom test suite.

The custom command is then the name of a script which contains `kubectl exec csi-hostpathplugin-0 mkdir -p "$@"`. That'll work because since https://github.com/kubernetes-csi/csi-driver-host-path/pull/14 the pod has a deterministic name.

Access to the `csi.sock` will be via ssh socket forwarding, set up outside of the sanity command invocation.

File: pkg/sanity/sanity.go (hunk 4 above)
Comment (darkowlzz) [reply pkg/sanity/sanity.go]: oops! thanks for noticing that.
I've added new checks in NodePublishVolume and NodeStageVolume to fail if the target paths don't exist. And adopted the above code snippet, this is actually better.

File: pkg/sanity/sanity.go (hunk 3 above)
Comment (darkowlzz) [reply pkg/sanity/sanity.go]: Sorry, that's really a mistake :sweat_smile: . Changed it.

**Comment by darkowlzz** [comment]
I've added command line support for passing external commands that can create the target path. An example usage is in the e2e test file.

I realized that it would be better if we change the names of a few flags and variables.
With block volume support, the target paths will no longer be directories always. The flags `mountdir` and `stagingdir` are file volume specific.
Maybe `mountpath` and `stagingpath` would be better names.
If we change this, we should also change the `Config` attributes that store custom target path creation callback functions `CreateTargetDir` and `CreateStagingDir` to `CreateTargetPath` and `CreateStagingPath`.

Maybe we need to main backwards compatibility?

File: pkg/sanity/sanity.go [hunk 5]
```diff
@@ -135,15 +158,23 @@ func (sc *SanityContext) setup() {
...
+	// If callback function for creating staging dir is specified, use it.
+	stagingPath, err := createMountTargetLocation(sc.Config.StagingPath, sc.Config.CreateStagingPathCmd, sc.Config.CreateStagingDir, sc.Config.CreatePathCmdTimeout)
+	Expect(err).NotTo(HaveOccurred(), "failed to create staging directory %s", stagingPath)
+	sc.stagingPath = stagingPath
 }
 
 func (sc *SanityContext) teardown() {
+	// Delete the created paths if any.
+	os.RemoveAll(sc.targetPath)
+	os.RemoveAll(sc.stagingPath)
```
darkowlzz: I added this in teardown because without this, some tests were using the directories created by the previous tests and that was resulting in the directory exists check in `createMountTargetLocation()` to be true and hence, not returning a target path even if it was created was passing, because that code block wasn't being executed.
With this, the directories are deleted for every test and not returning target path from `createMountTargetLocation()` results in failure.

pohly: While it's nicer if the test cleans up after itself, we then also need functions that correspond to `CreateStagingDir` and `CreateTargetDir`. The main use-case for these functions is for testing a driver that was deployed somewhere, and then those functions will not create the dirs locally.

pohly:
> Maybe mountpath and stagingpath would be better names.
> If we change this, we should also change the Config attributes that store custom target path creation callback functions CreateTargetDir and CreateStagingDir to CreateTargetPath and CreateStagingPath.

This is a good point. The proposed new names make more sense, but then the next release will have to be v2.0.0 due to the API change. I'm a bit undecided myself.

@lpabon what do you think?

pkg/sanity/sanity.go
darkowlzz: Right! I wasn't thinking in that way. Thanks, I'll add the corresponding cleanup options as well.

darkowlzz:
Added support for custom target path removal callback function and command line flag. The same is being used in the e2e test.

PR #160, opened 2019-01-30
 make mock driver support volume expand feature

mlmhl:
The volume resize feature has been already merged into the CSI spec, this PR makes the mock driver implements `ControllerExpandVolume` and `NodeExpandVolume` method so that we can use it to test the external-resizer.


gnufied:
We probably also need online/offline capabilities added to identity service but I think we can do this as a follow up.

mock/service/service.go
```diff
@@ -27,9 +27,10 @@ var Manifest = map[string]string{
 }
 
 type Config struct {
-	DisableAttach bool
-	DriverName    string
-	AttachLimit   int64
+	DisableAttach         bool
+	DriverName            string
+	AttachLimit           int64
+	NodeExpansionRequired bool
```
msau42: Do you also want ControllerExpansion to be configurable?  In case you want to do tests against a driver that doesn't support expansion?

gnufied: I think we could but can we do this as a follow up? I would also like to make online and offline configurable.

mlmhl:
@gnufied @msau42   Sorry for the late, I will solve these issues in a follow up PR.

PR #162, opened 2019-02-04
Use real volumes when testing no capabilities

### LINKED ISSUE #148: ValidateVolumeCapabilities Fails Because It Calls Non-existing Volume
Reproduction:
v1.1.0 branch
```
func TestDriver(t *testing.T) {
	go func() {
		d := getDriver(t)
		if err := d.Run(); err != nil {
			t.Fatal(err)
		}
	}()
	sc := &sanity.Config{
		TargetPath:  "/mnt/my-publish",
		StagingPath: "/mnt/my-staging",
		Address:     Endpoint,
	}
	sanity.Test(t, sc)
}
```
All tests pass except ValidateVolumeCapabilities which fails for the below reason.
```
STEP: reusing connection to CSI driver at unix:///tmp/test-csi.sock
STEP: creating a single node writer volume
STEP: validating volume capabilities
cleanup: deleting sanity-controller-validate-053C02BC-18C78554 = CSI-sanity-controller-validate-053C02BC-18C78554

------------------------------
• Failure [2.087 seconds]
Controller Service
/home/ubuntu/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
  ValidateVolumeCapabilities
  /home/ubuntu/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:591
    should return appropriate values (no optional values added) [It]
    /home/ubuntu/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:618

    Expected
        <[]*csi.VolumeCapability | len:0, cap:0>: nil
    not to be empty

    /home/ubuntu/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:671
------------------------------
2018-12-05T18:12:46Z INFO GRPC -- request: /csi.v1.Controller/ValidateVolumeCapabilities -- volume_id:"some-vol-id" volume_capabilities:<mount:<> access_mode:<mode:SINGLE_NODE_WRITER > > 
```
You can see that the volume created for this test was ``sanity-controller-validate-053C02BC-18C78554``, but the volume called for the test in the ValidateVolumeCapabilities step is ``some-vol-id``.

For some reason, the original volume-id isn't being used during subsequent steps of the test.  My ValidateVolumeCapabilities function is returning an error because the volume_id is invalid.

bswartz:
Negative tests that expect the plugin to return InvalidArgument because
no capabilities were supplied must use a real volume, otherwise plugins
might return NotFound instead.

Fixes #148

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -603,16 +603,56 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 
 		It("should fail when no volume capabilities are provided", func() {
 
-			_, err := c.ValidateVolumeCapabilities(
+			// Create Volume First
+			By("creating a single node writer volume")
+			name := uniqueString("sanity-controller-validate-nocaps")
+
+			vol, err := c.CreateVolume(
+				context.Background(),
+				&csi.CreateVolumeRequest{
+					Name: name,
+					VolumeCapabilities: []*csi.VolumeCapability{
+						{
+							AccessType: &csi.VolumeCapability_Mount{
+								Mount: &csi.VolumeCapability_MountVolume{},
+							},
+							AccessMode: &csi.VolumeCapability_AccessMode{
+								Mode: csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER,
+							},
+						},
+					},
+					Secrets:    sc.Secrets.CreateVolumeSecret,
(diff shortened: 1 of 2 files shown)


jsafrane:
@bswartz, would you mind fixing gofmt? Travis uses go version 1.10.3 and does not like your code.

Issue #146, opened 2018-11-26
Add support for generating a JUnit XML test report file


rodrigc:
It would be very handy if csi-sanity and other test suites can be enhanced to generate
a JUnit XML test report file:

https://onsi.github.io/ginkgo/#generating-junit-xml-output

This will help integrate with the Jenkins test report viewer

pohly:
You want this to be added to the csi-sanity command, right?

When defining your own test suite, you can already add the necessary code, nothing in csi-test must be modified for that.

/help-wanted

rodrigc:
@pohly Yes, my request is for this to be added to the csi-sanity command.

PR #164, opened 2019-02-11
Add '--csi.junitfile` flag

rodrigc:
This flag specifies a JUnit XML file which contains test results.

Fixes https://github.com/kubernetes-csi/csi-test/issues/146


cmd/csi-sanity/sanity_test.go
```diff
@@ -42,6 +42,7 @@ func init() {
 	flag.StringVar(&config.SecretsFile, prefix+"secrets", "", "CSI secrets file")
 	flag.Int64Var(&config.TestVolumeSize, prefix+"testvolumesize", sanity.DefTestVolumeSize, "Base volume size used for provisioned volumes")
 	flag.StringVar(&config.TestVolumeParametersFile, prefix+"testvolumeparameters", "", "YAML file of volume parameters for provisioned volumes")
+	flag.StringVar(&config.JUnitFile, prefix+"junitfile", "", "JUnit XML file containing test results")
```
pohly: Minor nit: the description makes and the commit message do not make it clear that this is going to be created as part of the test run.

I suggest "JUnit XML output file where test results will be written".

rodrigc: OK, I made the change

PR #156, opened 2019-01-24
release tools

pohly:
This syncs the build rules with the other kubernetes-csi repos.

PR #165, opened 2019-02-11
Add StartOnAddress function to listen on unix:// paths

jsafrane:


jsafrane:
cc @pohly, @msau42

driver/mock.go
```diff
@@ -61,6 +61,12 @@ func (m *MockCSIDriver) Start() error {
 	return nil
 }
 
+// StartOnAddress starts a new gRPC server listening on a random TCP loopback port.
+func (m *MockCSIDriver) Start() error {
```
pohly: Comment doesn't match the method.

I know it wasn't document before either (or at least not here), but when using random port allocation, how does the caller determine how to reach the mock driver, i.e. determine the port?

jsafrane: Good catch, comment is fixed now. Caller can call `MockCSIDriver.Address()` to get the actual address.

https://github.com/kubernetes-csi/external-attacher/blob/8f63c2c34b250355e73fa2d91e6cc65ea72f9645/pkg/connection/connection_test.go#L64-L66

jsafrane:
@lpabon, can you please release new version of csi-test? It blocks https://github.com/kubernetes-csi/external-attacher/pull/123

pohly:
I'd like to see PR #156 merged first. Then the new release becomes the first real-world test of image publishing with those build rules.

PR #179, opened 2019-03-19
Make mount and staging dir creation flexible

### LINKED ISSUE #144: sanity: more flexible mount and staging directory creation

pohly:
This change adds an option to attach callback functions to CSI sanity
for mount target directory creation. This could be used to customize the
target and staging mount path creation.

Also, adds targetPath and stagingPath fields to SanityContext to
store the actual target and staging paths, derived from the sanity
config.

Adds an e2e test section to test the above setup.

Fixes #144

This is PR #159 rebased on top of master to resolve a minor merge conflict.


pohly:
The original PR was merged.

PR #188, opened 2019-04-10
Switch to go modules

bswartz:


pohly:
Is this still using the "vendor" directory?

I ran "make test" in /tmp/csi-test and got lots of "finding" and "downloading" messages:
```
go: finding golang.org/x/exp v0.0.0-20190121172915-509febef88a4
go: finding golang.org/x/net v0.0.0-20180724234803-3673e40ba225
go: downloading google.golang.org/grpc v1.19.0
go: downloading github.com/container-storage-interface/spec v1.1.0
go: downloading github.com/onsi/ginkgo v1.8.0
go: downloading github.com/golang/protobuf v1.3.1
go: downloading gopkg.in/yaml.v2 v2.2.2
```

I think we still want to maintain and use the "vendor" directory, to avoid having to reach out to all those repos during a CI build.

Also this:
```
PASS
mock driver started
build cache is disabled by GOCACHE=off, but required as of Go 1.12
Makefile:30: recipe for target 'test-sanity' failed
make: *** [test-sanity] Error 1
```

bswartz:
> Is this still using the "vendor" directory?
> 
> I think we still want to maintain and use the "vendor" directory, to avoid having to reach out to all those repos during a CI build.

From what I understand, go modules obsoletes the use of vendor directories. Now instead there is a "module cache" located at $GOPATH/pkg/mod which serves the same purpose the vendor directory used to serve. The benefit of the module cache is that it can store more than 1 version of each package, and it can be stored outside your git repo (dramatically reducing the size of most git repos). CI builds rely on module caches/mirrors like Athena to avoid hitting the internet.

bswartz:
Thanks for reminding me that I forgot to delete the vendor directory with this change. That is done now.

msau42:
FYI, an overview of go modules: https://www.youtube.com/watch?v=OrMA16ASz2U

pohly:
Ben Swartzlander <notifications@github.com> writes:

>> Is this still using the "vendor" directory?
>> 
>> I think we still want to maintain and use the "vendor" directory, to avoid having to reach out to all those repos during a CI build.
>
> From what I understand, go modules obsoletes the use of vendor
> directories.

No, not necessarily:
https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away

"The initial series of vgo blog posts did propose dropping vendoring
entirely, but feedback from the community resulted in retaining support
for vendoring."

BTW, that FAQ also explains why I was seeing packages being downloaded:
one has to explicitly enable the use of the "vendor" directory with
-mod=vendor or GOFLAGS=-mod=vendor.

> The benefit of the module cache is that it can store more than 1
> version of each package,

That should also be true for the "vendor" directory.

> and it can be stored outside your git repo
> (dramatically reducing the size of most git repos).

Has that been a problem? Just wondering.

> CI builds rely on module caches/mirrors like Athena to avoid hitting
> the internet.

Do we have or will we have that when building in Prow?

bswartz:
> No, not necessarily: https://github.com/golang/go/wiki/Modules#how-do-i-use-vendoring-with-modules-is-vendoring-going-away "The initial series of vgo blog posts did propose dropping vendoring entirely, but feedback from the community resulted in retaining support for vendoring." BTW, that FAQ also explains why I was seeing packages being downloaded: one has to explicitly enable the use of the "vendor" directory with -mod=vendor or GOFLAGS=-mod=vendor.

Thanks for this helpful link. I didn't find that information in my reading about go modules.

>> The benefit of the module cache is that it can store more than 1 version of each package,
>> That should also be true for the "vendor" directory.
>> and it can be stored outside your git repo (dramatically reducing the size of most git repos).
> Has that been a problem? Just wondering.

I think that for most projects, the size of the vendor directory exceeds the size of the project's code by a factor of 10 or even 100. In practice this means that you'll hit any scaling problems 10-100x sooner than you otherwise would. For small projects this doesn't matter much, but I still don't like being wasteful.

>> CI builds rely on module caches/mirrors like Athena to avoid hitting the internet.
> Do we have or will we have that when building in Prow?

That's a good question. Since k8s is switching to go modules fairly aggressively I assume some kind of solution is in place, but I haven't looked at the specifics.

I will read up on how to keep the vendored stuff because I'd like to add go module support with minimal disruptions.

pohly:
Ben Swartzlander <notifications@github.com> writes:
>>> CI builds rely on module caches/mirrors like Athena to avoid hitting the internet.
>> Do we have or will we have that when building in Prow?
>
> That's a good question. Since k8s is switching to go modules fairly
> aggressively I assume some kind of solution is in place, but I haven't
> looked at the specifics.

I asked on Slack. Catherine replied: "Currently everyone using modules
in kubernetes-land is still vendoring, I think. We do not currently have
a cache."
(https://kubernetes.slack.com/archives/C09QZ4DQB/p1554994915357100)

Ben and Steve confirmed that there is currently no cache.

Let's keep vendoring the files until we know for sure that dropping them
won't have a negative impact on the CI builds. CI builds happen in
$GOPATH with go 1.12, so they should be using the "vendor" dir
automatically.

Note that we have a CI test which runs "test-vendor". The build rules
for that currently assume that "dep" is used, and the check runs
entirely locally.

Writing a similar check for "go mod" will be a bit tricky. I don't know
whether "go mod verify" can be used without actually reaching out to all
repos. If so, then that would negate the purpose of having the vendor
directory.

bswartz:
I'll fix this. Thanks for the additional info

bswartz:
@pohly PTAL

pohly:
You need to use a different trick for turning of test result caching:
```
diff --git a/hack/e2e.sh b/hack/e2e.sh
index c182b1d..e3a8c6c 100755
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -54,8 +54,8 @@ runTestAPI()
 	local pid=$!
         trap 'cleanup $pid $1' EXIT
 
-	GOCACHE=off go test -v ./hack/_apitest/api_test.go && \
-	GOCACHE=off go test -v ./hack/_embedded/embedded_test.go
+	go test -count=1 -v ./hack/_apitest/api_test.go && \
+	go test -count=1 -v ./hack/_embedded/embedded_test.go
 )
 
 runTestAPIWithCustomTargetPaths()
@@ -66,7 +66,7 @@ runTestAPIWithCustomTargetPaths()
 
 	# Running a specific test to verify that the custom target paths are called
 	# a deterministic number of times.
-	GOCACHE=off go test -v ./hack/_apitest2/api_test.go -ginkgo.focus="NodePublishVolume"
+	go test -count=1 -v ./hack/_apitest2/api_test.go -ginkgo.focus="NodePublishVolume"
 )
 
 runTestWithCustomTargetPaths()
```

But even then the `test-vendor` test will fail. You need to rebase or merge with master to get that test. Or we can run it in the CI:

/test pull-kubernetes-csi-csi-test

pohly:
You can try out a different implementation of the `test-vendor` check in this PR. That'll trigger a `test-subtree` failure, which will prevent accidentally merging that modified PR. The actual `test-vendor` change then has to come in via `csi-release-tools`. `release-tools/README.md` has a command for pushing the modification into a `csi-release-tools` fork for submission as a PR.

pohly:
During the transition period, `test-vendor` will have to be done so that it checks whether repo uses `dep` or `go mod` (can be done by checking for `Gopkg.toml`) and then do the right version of the check.

pohly:
I checked that your branch can still be pulled into a project that uses `dep` for vendoring. It might not be using the dependency information, though.

pohly:
Oops, the Prow jobs do not currently support the `test-vendor` check:
```
### test-vendor:
skipping check, dep >= 0.5 required
```

The build is passing because we still use Go 1.11.

Can you bump Go to 1.12.1 in `release-tools/travis.yml`?

bswartz:
@pohly I'm back in the office today and I'll update this PR based on your comments.

Makefile
```diff
@@ -16,6 +16,7 @@
 # but csi-sanity has its own build rules and only mock-driver gets
 # published as a container image.
 CMDS=mock-driver
+export GOFLAGS=-mod=vendor
```
pohly: Strictly speaking, we currently don't need to set `GOFLAGS=-mod=vendor` because when building in `GOPATH`, go 1.12 still defaults to non-module builds and then uses `vendor`. I checked with strace. But I agree that it makes sense to already add `GOFLAGS=-mod=vendor` because then we won't forget to add it when updating to 1.13 where modules are on by default.

How to do it is a bit trickier. The way it is done here overrides whatever `GOFLAGS` the developer may have in his environment. `make GOFLAGS=...` can be used to override this default.

I think I prefer to not set GOFLAGS for developers. Just set it for CI builds, by adding it to `release-tools/travis.yml` and `release-tools/prow.sh`.

hoyho:
>  test-vendor: 
>  	@ echo; echo "### $@:" 
>  	@ case "$$(dep version 2>/dev/null | grep 'version *:')" in \ 
>  		*v0.[56789]*) dep check && echo "vendor up-to-date" || false;; \ 
>  		*) echo "skipping check, dep >= 0.5 required";; \

Can we add a condition before `case "$$(dep version 2>/dev/null | grep 'version *:')" in \ 
 		*v0.[56789]*) ` to check if go mod is enable and Gopkg.toml is not exist. Then do nothing to avoid error but also keep the test-vendor script which other repo may need it.

pohly:
> Can we add a condition before case "$$(dep version 2>/dev/null | grep 'version *:')" in \ *v0.[56789]*) to check if go mod is enable and Gopkg.toml is not exist. Then do nothing to avoid error but also keep the test-vendor script which other repo may need it.

Suppressing the check is easy. What's hard is coming up with something similar for repos that use `go mod`. We do want to keep vendoring the source code for now (see previous comments), so we need a way to verify that dependency changes are reflected also in the vendored source code.

pohly:
> Would one way to check be try loading all the pkgs and see if that results in a diff?

If we do that for each PR, then it negates the purpose of having the sources vendored, because each build then needs to access all remote repos to do the checking.

We could try to skip it in Prow under certain specific situations:
- not handling a PR or
- the fabricated merge commit leaves go.mod and go.sum unchanged

For normal developers, `make test` will always do the check, so they are warned if they forget to vendor. We can include a message that tells them what to do, too.

Sounds doable, albeit a bit complex.

pohly:
Here's a modified test-vendor check which implements that idea:
```
test-vendor:
	@ echo; echo "### $@:"
	@ if [ -f Gopkg.toml ]; then \
		echo "Repo uses 'dep' for vendoring."; \
		case "$$(dep version 2>/dev/null | grep 'version *:')" in \
			*v0.[56789]*) dep check && echo "vendor up-to-date" || false;; \
			*) echo "skipping check, dep >= 0.5 required";; \
		esac; \
	  else \
		echo "Repo uses 'go mod' for vendoring."; \
		if [ "$${JOB_NAME}" ] && \
                   ( [ "$${JOB_TYPE}" != "presubmit" ] || \
                     [ $$(git diff "${PULL_BASE_SHA}..HEAD" -- go.mod go.sum vendor release-tools | wc -l) -eq 0 ] ); then \
			echo "Skipping vendor check because the Prow pre-submit job does not change vendoring."; \
		elif ! GO111MODULE=on go mod vendor; then \
			echo "ERROR: vendor check failed."; \
			false; \
		elif [ $$(git status --porcelain=2 -- vendor | wc -l) -gt 0 ]; then \
			echo "ERROR: vendor directory *not* up-to-date, it did get modified by 'GO111MODULE=on go mod vendor':"; \
			git status -- vendor; \
			git diff -- vendor; \
			false; \
		fi; \
	 fi;
```

However, when I try that with go 1.12.1, I do get a test failure:
```
### test-vendor:
Repo uses 'go mod' for vendoring.
ERROR: vendor directory *not* up-to-date, it did get modified by 'GO111MODULE=on go mod vendor':
On branch go-modules
Your branch is up-to-date with 'NetApp/go-modules'.
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   vendor/modules.txt

no changes added to commit (use "git add" and/or "git commit -a")
diff --git a/vendor/modules.txt b/vendor/modules.txt
index 49f67fc..13b15ec 100644
--- a/vendor/modules.txt
+++ b/vendor/modules.txt
@@ -95,6 +95,7 @@ google.golang.org/grpc/codes
 google.golang.org/grpc/reflection
 google.golang.org/grpc/status
 google.golang.org/grpc/connectivity
+google.golang.org/grpc/keepalive
 google.golang.org/grpc/balancer
 google.golang.org/grpc/balancer/roundrobin
 google.golang.org/grpc/credentials
@@ -109,7 +110,6 @@ google.golang.org/grpc/internal/envconfig
 google.golang.org/grpc/internal/grpcrand
 google.golang.org/grpc/internal/grpcsync
 google.golang.org/grpc/internal/transport
-google.golang.org/grpc/keepalive
 google.golang.org/grpc/metadata
 google.golang.org/grpc/naming
 google.golang.org/grpc/peer
release-tools/build.make:124: recipe for target 'test-vendor' failed
make: *** [test-vendor] Error 1
```

My version of go wrote the entries in a different order. I therefore only allow the test to be skipped if `release-tools` is also unmodified, otherwise we might end up updating Go without running the test.

What do you think? Shall we put that into `release-tools` from where @bswartz can pull it into this PR?

pohly:
@bswartz which version of go did you use to create the vendor directory?

I am a bit worried that re-doing the vendoring with 1.12.1 leads to a small difference (see above).

bswartz:
> @bswartz which version of go did you use to create the vendor directory?
> 
> I am a bit worried that re-doing the vendoring with 1.12.1 leads to a small difference (see above).

I can't remember which go version I used. I currently have 1.12.4, and I'm going to switch to 1.12.5. I'll do what you suggest in your latest comment though.

bswartz:
@pohly PTAL

pohly:
`make test-vendor` passed without problems now for me and in TravisCI. I used go 1.12.1, so it looks like any 1.12.x version is fine when checking or updating dependencies (one, two, all, right? :smile:).

/hold

Before merging let's have a chat how we'll roll this out to more repos.

pohly:
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-test/188/pull-kubernetes-csi-csi-test/1129120593698885632 shows that the `test-vendor` test ran, as intended.

pohly:
Discussed today in the CSI WG. The conclusion was to first finish the sidecar releases for 1.15 and do more experiments with "go mod" support in other repos (like csi-driver-host-path and csi-lib-utils).

One question which needs to be answered: what happens if a repo switches to "go mod" and some other repo which still uses "dep" pulls that repo in? Will "dep" be able to apply the constraints specified in `go mod`?

pohly:
Looks like Kubernetes is going to rely on proxy.golang.org for caching dependencies: https://github.com/kubernetes/test-infra/issues/13645

This means we could stop vendoring source code and just rely on that same cache?

msau42:
Sounds fine to me. We just need to set the GOPROXY env and update to go 1.13?

pohly:
Michelle Au <notifications@github.com> writes:
> Sounds fine to me. We just need to set the GOPROXY env and update to
> go 1.13?

The GOPROXY env var is only needed for go 1.12, 1.13 will enable it by
default.

If I read the test-infra issue correctly, all (?) Prow jobs will
automatically get GOPROXY set, so we might not have to do anything.

bswartz:
I'm sure someone has already thought about this, but using a proxy like that will dramatically increase reliance on network resources at build time compared to the vendoring approach. I'm surprised there's no plan to at least try to use a proxy that will be local to the build machines.

pohly:
> I'm surprised there's no plan to at least try to use a proxy that will be local to the build machines.

Have a look at https://github.com/kubernetes/test-infra/issues/13645. It has been discussed and the conclusion was that for all practical purposes, proxy.golang.org *is* local - it's apparently all hosted on Google Cloud. If that changes, then setting GOPROXY differently can still take care of that once the need arises.

bswartz:
@pohly Thanks I forgot that golang is actually a google project and so it's unsurprising the proxy is hosted in the same place as the kubernetes build infra. Now I just need to figure out how to run my own proxy for my own builds.

bswartz:
https://github.com/kubernetes-csi/csi-test/pull/224

PR #195, opened 2019-04-26
sanity: capture newly created volume IDs

avalluri:
In one of ListVolumes tests, newly created volume ids were not captured. Though the test has needed cleaning up code, it was affected because of the empty newVolIDs array. This leads to leaking of volumes.

 /kind bug

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -281,8 +281,8 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 			// currentTotalVols is the total number of volumes at a given time. It
 			// is used to verify that all the volumes have been listed.
 			currentTotalVols := 0
-			// newVolIDs to keep a record of the newly created volume ids.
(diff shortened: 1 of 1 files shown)


pohly:
Thanks for highlighting this issue. However, the issue is even more severe: if one of the asserts fails after creating the volume, the cleanup code won't get executed. I missed that when merging https://github.com/kubernetes-csi/csi-test/pull/80.

Can you change the test such that it uses https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/cleanup.go?

avalluri:
> Thanks for highlighting this issue. However, the issue is even more severe: if one of the asserts fails after creating the volume, the cleanup code won't get executed. I missed that when merging #80.
> Can you change the test such that it uses https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/cleanup.go?

@pohly, I updated the PR to use Cleanup interface as you suggested.

PR #200, opened 2019-05-06
CreateVolume respects the topology requirements of the node

### LINKED ISSUE #194: csi sanity "Node Service"."should work" test seems to be more restrictive than the spec
CreateVolumeRequest in the test does not containt AccessibilityRequirements
https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/node.go#L632
which is allowed in the spec:
https://github.com/container-storage-interface/spec/blob/master/lib/go/csi/csi.pb.go#L833
However, in this case "the SP MAY choose where the provisioned volume is accessible from" rather than required to. 

If there's no universally accessible disk provided by the SP the most logical thing to do seems to fail the CreateVolume which will fail the test:
https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/node.go#L648

One might suggest that SP choose any of the available zones and provide a volume, but in this case the test might fail further at
https://github.com/kubernetes-csi/csi-test/blob/master/pkg/sanity/node.go#L684
since the node that the driver runs on might well end up in the different accessibility zone which will make publishing the volume on the node impossible

alexanderKhaustov:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #194 

```release-note
NONE
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -1151,8 +1151,8 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 							},
 						},
 					},
-					Secrets:    sc.Secrets.CreateVolumeSecret,
-					Parameters: sc.Config.TestVolumeParameters,
+					Secrets:                   sc.Secrets.CreateVolumeSecret,
+					Parameters:                sc.Config.TestVolumeParameters,
 					AccessibilityRequirements: accReqs,
 				},
 			)
@@ -1514,8 +1514,8 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 							},
 						},
 					},
-					Secrets:    sc.Secrets.CreateVolumeSecret,
-					Parameters: sc.Config.TestVolumeParameters,
+					Secrets:                   sc.Secrets.CreateVolumeSecret,
+					Parameters:                sc.Config.TestVolumeParameters,
 					AccessibilityRequirements: accReqs,
 				},
 			)
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -1116,6 +1116,8 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 			Expect(serverError.Code()).To(Equal(codes.InvalidArgument))
 		})
 
+		// CSI spec poses no specific requirements for the cluster/storage setups that a SP MUST support. To perform
+		// meaningful checks the following test assumes that topology-aware provisioning on a single node setup is supported
 		It("should return appropriate values (no optional values added)", func() {
 
 			By("getting node information")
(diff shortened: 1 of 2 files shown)


pkg/sanity/controller.go
```diff
@@ -1118,6 +1118,21 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
...
+				context.Background(),
+				&csi.NodeGetInfoRequest{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(ni).NotTo(BeNil())
+			Expect(ni.GetNodeId()).NotTo(BeEmpty())
+
+			var accReqs *csi.TopologyRequirement
+			if ni.AccessibleTopology != nil {
+				accReqs = &csi.TopologyRequirement{
+					Requisite: []*csi.Topology{ni.AccessibleTopology},
```
msau42: This assumes that the storage backend can support a one-node topology, but may not be the case for something that requires say a min of 3 nodes for replication.

I'm unsure what would be a great way to support that in the test though.  This is probably ok for now, but it might be good to add a comment about the assumption.

pohly: csi-sanity essentially simulates a CO with a single node, so I think this code change is consistent with the rest of the testing.

@alexanderKhaustov can you add a short comment here that explains that?

alexanderKhaustov: Sorry for the delay
I'm not sure I understand the potentially problematic scenario.
The SP is returning topology information for a node, but then it doesn't support provisioning a volume with this topology, because of a requirement (existing only externally to csi plugin/spec?) that a volume can only be created if there are no less than 3 nodes with some specific topology?
If that is the case then it seems that such an external peculiarity is hardly within the scope of csi-test, is it not? (csi-test aside, it seems that such a plugin would also perform unexpectedly for a CO end-user, would it not?)

pohly: @alexanderKhaustov: I agree. Let me try to propose a comment that you can put above the new code and then if @msau42 has no additional comments we can merge it:

> csi-sanity testing emulates the behavior of a CO with a single node. The name and topology of that node is provided by the driver itself. Here we ensure that the new volume gets provisioned on that node, because otherwise staging it on the node later in the test might fail.

msau42: I think it would be good to add a note that this test does not support storage backends that require more than 1 node.

alexanderKhaustov: I've checked the spec more thoroughly, seems like there are no requirements for the setups that a SP MUST support. So I added comments about test assumptions.

As an aside, with regard to the topology returned by NodeGetInfo the spec suprisingly states that 
> COs MAY use this [topology] information ... when scheduling workloads

(https://github.com/container-storage-interface/spec/blob/master/spec.md#nodegetinfo)
rather than MUST use which seems more reasonable. Do you know any reason for that?

pohly: I don't know why topology information is allowed to be ignored.

msau42:
/approve

Thanks!

PR #205, opened 2019-06-28
Closes #204 - Treat list volume tokens as opaque

### LINKED ISSUE #204: csi-sanity incorrectly testing vol list tokens
CSI-sanity is making incorrect assumptions about the tokens generated by CSI plugins to do volume listing pagination, so it incorrectly fails on some plugins.

Existing test `should return next token when a limited number of entries are requested` **assumes that `next_token` is an integer** and that it is equivalent to the number of volumes listed:
```
Expect(nextToken).To(Equal(strconv.Itoa(maxEntries)))
```

But that is not what the CSI spec says.  In the spec `next_token` and `starting_token` are strings, and as it is with all pagination mechanisms, these are opaque to the caller.  The caller must just blindly put the data from `next_token` into `starting_token` as its contents will be plugin specific.

List pagination of plugins must support deletion and creation of resources between page requests, but we are not testing this either.

Akrog:
**What type of PR is this?**
/kind bug

When listing volumes using pagination we must treat the "next_token"
field as opaque data and not make assumptions about what's in it.

This patch removes tests that assume tokens are integers and that they
represent the number of volumes that have been listed, replacing one of
them with one to check that pagination works as expected when new
volumes are created and old ones deleted between page requests.

Fixes #204 

```release-note
Remove volume listing tests that assumed tokens were integers and replace them with one that confirms pagination works when adding/deleting volumes between page requests.  
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -197,32 +197,7 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 			Expect(serverError.Code()).To(Equal(codes.Aborted))
 		})
 
-		It("should fail when the starting_token is greater than total number of vols", func() {
-			// Get total number of volumes.
-			vols, err := c.ListVolumes(
-				context.Background(),
-				&csi.ListVolumesRequest{})
-			Expect(err).NotTo(HaveOccurred())
-			Expect(vols).NotTo(BeNil())
-
-			totalVols := len(vols.GetEntries())
-
-			// Send starting_token that is greater than the total number of volumes.
-			vols, err = c.ListVolumes(
-				context.Background(),
-				&csi.ListVolumesRequest{
-					StartingToken: strconv.Itoa(totalVols + 5),
-				},
-			)
-			Expect(err).To(HaveOccurred())
-			Expect(vols).To(BeNil())
-
-			serverError, ok := status.FromError(err)
-			Expect(ok).To(BeTrue())
-			Expect(serverError.Code()).To(Equal(codes.Aborted))
-		})
-
-		It("check the presence of new volumes in the volume list", func() {
+		It("check the presence of new volumes and absence of deleted ones in the volume list", func() {
 			// List Volumes before creating new volume.
 			vols, err := c.ListVolumes(
 				context.Background(),
@@ -284,17 +259,14 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 			Expect(len(vols.GetEntries())).To(Equal(totalVols))
 		})
 
-		It("should return next token when a limited number of entries are requested", func() {
+		It("pagination should detect volumes added between pages and accept tokens when the last volume from a page is deleted", func() {
 			// minVolCount is the minimum number of volumes expected to exist,
 			// based on which paginated volume listing is performed.
-			minVolCount := 5
+			minVolCount := 3
 			// maxEntried is the maximum entries in list volume request.
 			maxEntries := 2
-			// currentTotalVols is the total number of volumes at a given time. It
-			// is used to verify that all the volumes have been listed.
-			currentTotalVols := 0
-			// newVols to keep a record of the newly created volume names and ids.
-			newVols := map[string]string{}
+			// existing_vols to keep a record of the volumes that should exist
+			existing_vols := map[string]bool{}
 
 			// Get the number of existing volumes.
 			vols, err := c.ListVolumes(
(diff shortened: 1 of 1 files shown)


Akrog:
This [CI failure](https://travis-ci.org/kubernetes-csi/csi-test/builds/551713193#L777) is caused by a deficient CSI pagination implementation, as it doesn't play well with volumes being created/deleted in between page requests.

I'll relax the test, as there can be plugins out there with this kind of implementation and they don't go *strictly speaking* against the current CSI spec.  Though, in my opinion doing a listing and not getting all existing volumes is a bad thing.

pohly:
> This CI failure is caused by a deficient CSI pagination implementation, as it doesn't play well with volumes being created/deleted in between page requests.
> 
> I'll relax the test, as there can be plugins out there with this kind of implementation and they don't go strictly speaking against the current CSI spec. Though, in my opinion doing a listing and not getting all existing volumes is a bad thing.

The spec explicitly allows missing volumes: "If volumes are created and/or deleted while the CO is concurrently paging through ListVolumes results then it is possible that the CO MAY either witness duplicate volumes in the list, not witness existing volumes, or both." So yes, the test had to be relaxed.

At first glance it looks like it would be nice if the result was consistent. However, that makes the pagination support more complex to implement, so perhaps that's why the spec authors decided to not require it.

/ok-to-test

Akrog:
> > This CI failure is caused by a deficient CSI pagination implementation, as it doesn't play well with volumes being created/deleted in between page requests.
> > I'll relax the test, as there can be plugins out there with this kind of implementation and they don't go strictly speaking against the current CSI spec. Though, in my opinion doing a listing and not getting all existing volumes is a bad thing.
> 
> The spec explicitly allows missing volumes: "If volumes are created and/or deleted while the CO is concurrently paging through ListVolumes results then it is possible that the CO MAY either witness duplicate volumes in the list, not witness existing volumes, or both." So yes, the test had to be relaxed.
> 
> At first glance it looks like it would be nice if the result was consistent. However, that makes the pagination support more complex to implement, so perhaps that's why the spec authors decided to not require it.
> 
> /ok-to-test

My bad, I had completely forgotten about that part of the spec.

In my opinion those behaviors effectively turn pagination into something useless.  In a running environment it may happen that there are volumes you never get to see if you use listings with pagination.

But that is a completely different conversation unrelated to this project and the testing of CSI plugins.  ;-)

pohly:
Gorka Eguileor <notifications@github.com> writes:
> In my opinion those behaviors effectively turn pagination into
> something useless.  In a running environment it may happen that there
> are volumes you never get to see if you use listings with pagination.

Even without pagination a CO may miss volumes that get created shortly
after issuing the ListVolumes call. I suppose it just has to do that at
regular intervals, then eventually it'll see all volumes.

pkg/sanity/controller.go
```diff
@@ -197,32 +197,7 @@ var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
 			Expect(serverError.Code()).To(Equal(codes.Aborted))
 		})
 
-		It("should fail when the starting_token is greater than total number of vols", func() {
```
pohly: Why was your driver failing this test? `<totalVols + 5>` isn't a valid starting token, or is it?

Would it be better to pass a random UUID?

Akrog: This test is incorrect because the CSI spec defines tokens as strings, and these strings may not be representations of integers (spec doesn't say anything about the token format, so these are plugin specific).

And even *if* they were integers, they may not indicate the numbers of created volumes. For example they could be integers representing a timestamp.

So the only thing we can tell about tokens is that a weird string will probably fail (we test this in the previous test), and that a token returned by the plugin must be acceptable and not fail when passes as starting point for the next pagination request, even if the volumes have changed.

msau42:
There's always going to be races between List and Create/Delete.  A reconciling architecture helps address this and avoids complex serialization/coordination.

pkg/sanity/controller.go
pohly: > This test is incorrect because the CSI spec defines tokens as strings, and these strings may not be representations of integers

But the spec also doesn't say that the token must not be the representation of integers. The wording of the test isn't good, but the check itself makes sense.

The test should be called "should fail when the starting_token is invalid", and then we need a good way of creating an invalid token.

Akrog: That is correct, the spec doesn't say that the token cannot be an integer.  In Ember-CSI our tokens are integers representing the volume creation time so we can use them to filter volumes and make sure we don't return duplicates and always return newer volumes.

The test you mention already exists, it's the one before this one, L185-L198.  Although it is possible that "invalid-token" could be a valid token for some plugin.

Akrog:
> There's always going to be races between List and Create/Delete. A reconciling architecture helps address this and avoids complex serialization/coordination.

Yes, but there are ways to minimize the impact of these, for example ordering your volumes by creation time and returning a token with the time of the last volume returned.  That way you won't return duplicates on pagination and you will automatically return newer volumes.

A driver that uses integers that represent the number of volumes returned as tokens is a pretty bad implementation, as it is subject to both returning duplicates (depending on the order used to list the volumes) and missing volumes.

With a big enough number of volumes and enough create/delete operations you could make many consecutive complete paginated list requests and still not have seen some of the existing volumes in all the complete lists you have obtained.  But this is a plugin problem, since the spec allows it, and the specific plugin will get the bug if that happens.

pkg/sanity/controller.go
pohly: > The test you mention already exists, it's the one before this one

Oh, right. In that case I agree, "should fail when the starting_token is greater than total number of vols" can be removed. It might be a relevant test case for the CSI hostpath driver, but that's no reason to have it here.

The risk is too high that the integer actually is treated as a valid token by some driver, which then fails the test. Is that what happened with your driver?

Akrog: Exactly, any integer number is a valid token for Ember-CSI.

PR #208, opened 2019-07-18
Tolerate nil capabilities for plugins that don't provide CONTROLLER_SERVICE

wongma7:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**: providing CONTROLLER_SERVICE is optional. Node tests should not fail if it's not provided. With these changes, it's possible to do
$ csi-sanity --csi.endpoint=./csi.sock -ginkgo.skip='\[Controller.Server\]'
to test a node-only plugin against node-specific tests

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -92,7 +92,7 @@ func isControllerCapabilitySupported(
 	return false
 }
 
-var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
+var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *SanityContext) {
 	var (
 		c csi.ControllerClient
 		n csi.NodeClient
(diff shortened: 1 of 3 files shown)


pkg/sanity/controller.go
```diff
@@ -92,7 +92,7 @@ func isControllerCapabilitySupported(
 	return false
 }
 
-var _ = DescribeSanity("Controller Service", func(sc *SanityContext) {
+var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *SanityContext) {
```
pohly: That looks rather odd. Can't you simply skip based on the existing "Controller Service" string?

wongma7: yes, agreed it's redundant, but I think it's neater if everything has the same tag.

pohly: Ah, I didn't know that this tag is already in use elsewhere. Okay, then it makes sense.

PR #209, opened 2019-07-19
ephemeral volumes

pohly:
**What type of PR is this?**
/kind feature

This enables testing of recent csi-mock with recent Kubernetes by allowing the Kubernetes tests to make the TargetPath existence check more permissive and the testing of ephemeral inline volumes in Kubernetes.

All of the "CSI Mock volume" tests (https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/csi_mock_volume.go#L59) still pass with this new driver.

```release-note
csi-mock driver now supports relaxing the target path check (--permissive-target-path) and ephemeral inline volumes
```


lpabon:
Looks good

/lgtm
/approve

PR #210, opened 2019-07-19
cleanup fix + enhancement

pohly:
**What type of PR is this?**
/kind feature

Cleaning up may have leaked volumes due to an incorrect target path. Stress testing may use multiple goroutines, and then data structures in the cleanup code must be protected.

The concurrency support was developed for a stress test in PMEM-CSI. That's also were we noticed the leaked volumes.

```release-note
the sanity volume code can be used concurrently and avoids a potential volume leak because of an incorrect target path
```

### DIFF
--- a/pkg/sanity/cleanup.go
+++ b/pkg/sanity/cleanup.go
@@ -19,6 +19,7 @@ package sanity
 import (
 	"context"
 	"log"
+	"sync"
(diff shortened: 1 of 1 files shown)


lpabon:
Nice

/lgtm
/approve

PR #212, opened 2019-08-06
Add ID Generator interface for generating valid IDs for Volume and Node

### LINKED ISSUE #103: Support volume name generate interface
There are some cloud provider like **Linode**, has some constraints on volume (like volume id is integer number). So, It will be helpful to add a `VolumeNameGenerate interface{}`, then pass it to the controller.

davidz627:
/kind api-change
/kind feature

Fixes #103

```release-note
Adds an interface for generating Volume/Node IDs that users can plug in for ID's that must conform to valid format
```


davidz627:
/cc @displague @tamalsaha @sanjid133
does this work for your issues

davidz627:
Tested and working with PD driver

msau42:
We'll need to cut a new major for this, which I think is fine.  Would like one more reviewer on this.

/assign @jsafrane @lpabon 
/approve

davidz627:
> We'll need to cut a new major for this, which I think is fine. Would like one more reviewer on this.

Why is that? If this config isn't set it will default to doing the same thing as before (except with UUID's instead of just using `fakevol`/`testvol`/`id` etc.)

msau42:
Oh I see, this is a new optional config argument. I think that's fine then.

pkg/sanity/util.go
```diff
@@ -0,0 +1,52 @@
...
+	GenerateInvalidNodeID() string
+}
+
+var _ IDGenerator = &DefaultIDGenerator{}
+
+type DefaultIDGenerator struct {
+}
+
+func (d DefaultIDGenerator) GenerateUniqueValidVolumeID() string {
+	return fmt.Sprintf("fake-volid-%s", uuid.New().String()[:10])
```
jsafrane: typo: `volid`

jsafrane: ah, it's not `valid`, it's `volumeID`, sorry for the noise

pkg/sanity/sanity.go
```diff
@@ -120,6 +120,8 @@ type Config struct {
 	RemoveStagingPathCmd string
 	// Timeout for the executed commands for path removal.
 	RemovePathCmdTimeout int
+
+	IDGen IDGenerator
```
jsafrane: Description (esp. of the default behavior) would be useful.

pkg/sanity/util.go
```diff
@@ -0,0 +1,52 @@
...
+
+package sanity
+
+import (
+	"fmt"
+
+	"github.com/google/uuid"
+)
+
+type IDGenerator interface {
```
jsafrane: Again, I am missing documentation (incl. the members)

davidz627:
sweet! @msau42 @lpabon have time to cut a new release?

PR #220, opened 2019-09-23
sanity: fix nil pointer error for IDGen when using Ginkgo suite

pohly:
**What type of PR is this?**

/kind bug

When adding IDGen, only one of the two code paths that trigger testing
was updated such that it sets a default ID generator when the config
doesn't already have one. GinkgoTest was not updated.

As a result, existing E2E suites which use GinkgoTest and don't set the
new field crash at runtime with a nil pointer error when updated to
csi-test 2.2.0.

Updating the instance provided by the caller is also a bit
questionable because it is an undocumented side effect. It's cleaner
to treat the struct as read-only and implement the fallback in an
accessor function.

We should create a 2.2.1 with this fix...

```release-note
fix nil pointer error when Config.IDGen is not set
```

### DIFF
--- a/pkg/sanity/sanity.go
+++ b/pkg/sanity/sanity.go
@@ -121,9 +121,10 @@ type Config struct {
 	// Timeout for the executed commands for path removal.
 	RemovePathCmdTimeout int
 
-	// IDGen is an optional interface for callers to provide a generator for
-	// valid Volume and Node IDs. Defaults to DefaultIDGenerator which generates
-	// generic string IDs
+	// IDGen is an optional interface for callers to provide a
+	// generator for valid Volume and Node IDs. If unset,
+	// it will be set to a DefaultIDGenerator instance when
+	// passing the config to Test or GinkgoTest.
 	IDGen IDGenerator
 }
 
@@ -143,6 +144,21 @@ type SanityContext struct {
 	StagingPath string
 }
 
+// newContext sets up sanity testing with a config supplied by the
+// user of the sanity package. Ownership of that config is shared
+// between the sanity package and the caller.
+func newContext(reqConfig *Config) *SanityContext {
+	// To avoid runtime if checks when using IDGen, a default
+	// is set here.
+	if reqConfig.IDGen == nil {
+		reqConfig.IDGen = &DefaultIDGenerator{}
+	}
+
+	return &SanityContext{
+		Config: reqConfig,
+	}
+}
+
 // Test will test the CSI driver at the specified address by
 // setting up a Ginkgo suite and running it.
 func Test(t *testing.T, reqConfig *Config) {
@@ -158,14 +174,7 @@ func Test(t *testing.T, reqConfig *Config) {
 		}
 	}
 
-	if reqConfig.IDGen == nil {
-		reqConfig.IDGen = &DefaultIDGenerator{}
-	}
-
-	sc := &SanityContext{
-		Config: reqConfig,
-	}
-
+	sc := newContext(reqConfig)
 	registerTestsInGinkgo(sc)
 	RegisterFailHandler(Fail)
 
@@ -180,11 +189,11 @@ func Test(t *testing.T, reqConfig *Config) {
 	}
 }
 
+// GinkoTest is another entry point for sanity testing: instead of directly
+// running tests like Test does, it merely registers the tests. This can
+// be used to embed sanity testing in a custom Ginkgo test suite.
 func GinkgoTest(reqConfig *Config) {
-	sc := &SanityContext{
-		Config: reqConfig,
-	}
-
+	sc := newContext(reqConfig)
 	registerTestsInGinkgo(sc)
 }
 


pohly:
/cc @okartau

Found by Olev when updating PMEM-CSI to csi-test 2.2.0.

pkg/sanity/sanity.go
```diff
@@ -127,6 +127,15 @@ type Config struct {
 	IDGen IDGenerator
 }
 
+// GetIDGen always returns a valid ID generator. If none is set in the configuration,
+// then the DefaultIDGenerator is returned.
+func (c Config) GetIDGen() IDGenerator {
+	if c.IDGen != nil {
+		return c.IDGen
+	}
+	return defaultIDGen
```
davidz627: defaulting in an accessor is a code smell. This will surprise someone eventually. Could you just provide a default ID generator in the GinkgoTest? If you want to fix it once and for all you could also make the creation of a `Config` into a `NewConfig` function which defaults `IDGen` if not specified.

pkg/sanity/util.go
```diff
@@ -44,7 +44,7 @@ type IDGenerator interface {
 	GenerateInvalidNodeID() string
 }
 
-var _ IDGenerator = &DefaultIDGenerator{}
+var defaultIDGen IDGenerator = &DefaultIDGenerator{}
```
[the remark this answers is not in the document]
pohly: Or just return `&DefaultIDGenerator{}` - but that becomes mute if we don't use the accessor at all.

pkg/sanity/sanity.go
pohly: A `NewConfig` function is useful and would simplify the addition of new entries were the null entry isn't a sensible default. But depending on it being used to construct the `Config` is an API break.

For now I'll go for a deep-copy function instead which takes the user supplied `Config` and sanitizes it, like you did in `Test`.

pohly: Copying `Config` doesn't work. That the tests use the user-supplied instance is part of the API, too. In PMEM-CSI we use that to fill in some fields that depend on the current test in a `BeforeEach`.

I guess we have to continue with updating the user-supplied instance.

pohly: https://github.com/pohly/csi-test/commit/8ac7c59f5ae6a3a35251d0fc2d237b9d230d1197 contains a version which updates `IDGen` in both code paths.

But I am not convinced that this is actually the better solution compared to the current one with the accessor. It creates a special case just for `IDGen`, everything else is checked at test runtime.

@davidz627 perhaps I can convince you that the accessor isn't so bad after all?  :neutral_face:
I updated the defaulting to not use a global variable.

davidz627: I'm still slightly confused as to why callers (or a constructor) can't just set the `IDGen` to a default when creating the `Config` struct? If the `Config` struct must have an `IDGen` field set to function (it does) it should enforce that at creation time.

Another issue with defaulting in the accessor is now you've created an "API break" in that you MUST use the accessor. It's easy to imagine a scenario where someone adds code that doesn't use the accessor and grabs the `IDGen` directly and breaks.

pohly: > I'm still slightly confused as to why callers

Because it's an API break. Callers who used to work with 2.1.0 crash with 2.2.0.

> (or a constructor)

Because there is none. It would be nice to have one for all those cases where the default isn't the null value, but again, introducing one now is an API break and would imply that we need to switch to 3.0.0.

> can't just set the IDGen to a default when creating the Config struct?  If the Config struct must have an IDGen field set to function (it does) it should enforce that at creation time.

The comment says "IDGen is an optional interface for callers", so according to that it is not required that IDGen is set.

> Another issue with defaulting in the accessor is now you've created an "API break" in that you MUST use the accessor.

That depends on how the API description in https://github.com/kubernetes-csi/csi-test/blob/5c24e9081426e7b6c628660063bac3787d7063b2/pkg/sanity/sanity.go#L124-L127 is interpreted: I understand it as "nil" is a valid value and thus anyone reading that field must check for it. The accessor is just a convenience function for that, it's not required.

davidz627: I still think defaulting `IDGen` either or creation or near creation of the `Config` object is the correct way to do this. There's no API break and theres no unexpected behavior around accessing it.

The comment about `IDGen` being `optional` is probably incorrect as we can see it's being assumed + used in several places in the tests. It is actually required and should be defaulted or set by a caller.

If for some reason this doesn't work I would prefer to rev the major version by requiring a constructor (honestly requiring a constructor should never be the case - we should be able to default elsewhere where the config is "initialized") with the defaulting rather than force usage of an accessor that has implicit defaulting behavior.

misterikkit: > introducing one [constructor] now is an API break

Adding a new function is generally not considered an API break. The accessor you created is also a new function, and I don't consider that to be an API break either. Both approaches fix the bug in a new API, and users will have to switch to that API to get the bug fix.


```
// IDGen is an optional interface for callers to provide a generator for
// valid Volume and Node IDs. Defaults to DefaultIDGenerator which generates
// generic string IDs
```

My interpretation is close to yours, @pohly. To me, this says "Users don't need to supply this value and anyone reading it should expect the default behavior provided by `DefaultIDGenerator`". Since users are expected to create this struct directly, that means the nil value should be a useful default. If users were using a constructor, I would assume the constructor to handle defaulting.

The fact that the intention was to have this default, but no defaulting was provided in the library is a bug. However, I feel the bug lies with the *API design* and not its implementation. i.e. Asking users to initialize a struct directly means that the nil value should be useful, but that is not possible for interface types.

If the impact on users is comparable (minor code change to use new API) then we should focus on which API is preferable. The options I see here are,

1. Add an accessor with defaulting behavior
2. Add a constructor with defaulting behavior
3. Document that users must explicitly provide `DefaultIDGenerator` or another generator.

Option 1 is the worst fit with the existing API. The Config type has no methods or accessors today. Defining one for a single field is likely to be missed by someone when they access every other field directly.

Option 2 provides the most forward compatibility, since a constructor function can usually adopt a [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis) pattern.

Option 3 is updating the comments to describe the API implemented rather than the one intended. This may be reasonable, given how few users this library has.

Regarding DeepCopy,
> That the tests use the user-supplied instance is part of the API, too.

If a behavior is observable, that does not mean it must be treated like an API. (see also, [Hyrum's Law](https://www.hyrumslaw.com)) While it is good to avoid breaking users of this library, it is also good to avoid accumulating tech debt.


Just as a side comment, the [godoc](https://godoc.org/github.com/kubernetes-csi/csi-test/pkg/sanity) for this package does not make it easy to figure out how to use this package. Is there documentation elsewhere that users should be reading? Should we link to it from the godoc?

pohly: >> introducing one [constructor] now is an API break
>
> Adding a new function is generally not considered an API break.

Yes, as long as calling it is optional. But the idea was to make calling
the constructor mandatory because there new non-null values (like the
interface implementation) would be set. That would be an API break,
because current users of the API don't call it and then fail.

> The fact that the intention was to have this default, but no defaulting was provided in the library is a bug. However, I feel the bug lies with the *API design* and not its implementation. i.e. Asking users to initialize a struct directly means that the nil value should be useful, but that is not possible for interface types.

I don't think there has been much thought spent on API design in this
package... I suppose we now try to reconstruct what the API really is
based on the exported types and how they get used in reality.

> If the impact on users is comparable (minor code change to use new API) then we should focus on which API is preferable. The options I see here are,
>
> 1. Add an accessor with defaulting behavior
> 2. Add a constructor with defaulting behavior
> 3. Document that users must explicitly provide `DefaultIDGenerator` or another generator.
>
> Option 1 is the worst fit with the existing API. The Config type has
> no methods or accessors today. Defining one for a single field is
> likely to be missed by someone when they access every other field
> directly.
>
> Option 2 provides the most forward compatibility, since a constructor
> function can usually adopt a [functional
> options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)
> pattern.

That would also be my preferred solution. But let's postpone that until
we need to break the API for some other reason.

As a short-term solution let's fix the missing initialization, which is
the alternative solution that I mentioned in https://github.com/kubernetes-csi/csi-test/pull/220#discussion_r328013614

I've updated this PR with that code.

> Option 3 is updating the comments to describe the API implemented
> rather than the one intended. This may be reasonable, given how few
> users this library has.
>
> Regarding DeepCopy,
>> That the tests use the user-supplied instance is part of the API, too.
>
> If a behavior is observable, that does not mean it must be treated like an API. (see also, [Hyrum's Law](https://www.hyrumslaw.com)) While it is good to avoid breaking users of this library, it is also good to avoid accumulating tech debt.

True. But this aspect of the package is useful, so it should be part of
the API.

> Just as a side comment, the
> [godoc](https://godoc.org/github.com/kubernetes-csi/csi-test/pkg/sanity)
> for this package does not make it easy to figure out how to use this
> package. Is there documentation elsewhere that users should be
> reading?

Unfortunately not. New users probably discover it based on usage in
other projects, which then also serves as example for how to use it.

misterikkit: Thanks for the quick reply.

> > > introducing one [constructor] now is an API break
> > 
> > 
> > Adding a new function is generally not considered an API break.
> 
> Yes, as long as calling it is optional. But the idea was to make calling
> the constructor mandatory because there new non-null values (like the
> interface implementation) would be set. That would be an API break,
> because current users of the API don't call it and then fail.

My point was that adding an accessor is also a new "mandatory" function. However, I think "mandatory" is not the only way to view such a change. Any client that continues to use the existing API will experience the exact same behavior, bug and all. From that perspective, this is not a breaking API change.

> 
> > The fact that the intention was to have this default, but no defaulting was provided in the library is a bug. However, I feel the bug lies with the _API design_ and not its implementation. i.e. Asking users to initialize a struct directly means that the nil value should be useful, but that is not possible for interface types.
> 
> I don't think there has been much thought spent on API design in this
> package... I suppose we now try to reconstruct what the API really is
> based on the exported types and how they get used in reality.

That's a fair assessment, but I do not believe we need to be absolutely rigid with regard to the existing API. The cost of moving all callers on to a new and improved API may be worth it for said API.

> 
> > If the impact on users is comparable (minor code change to use new API) then we should focus on which API is preferable. The options I see here are,
> > 
> > 1. Add an accessor with defaulting behavior
> > 2. Add a constructor with defaulting behavior
> > 3. Document that users must explicitly provide `DefaultIDGenerator` or another generator.
> > 
> > Option 1 is the worst fit with the existing API. The Config type has
> > no methods or accessors today. Defining one for a single field is
> > likely to be missed by someone when they access every other field
> > directly.
> > Option 2 provides the most forward compatibility, since a constructor
> > function can usually adopt a [functional
> > options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis)
> > pattern.
> 
> That would also be my preferred solution. But let's postpone that until
> we need to break the API for some other reason.
> 
> As a short-term solution let's fix the missing initialization, which is
> the alternative solution that I mentioned in [#220 (comment)](https://github.com/kubernetes-csi/csi-test/pull/220#discussion_r328013614)
> 
> I've updated this PR with that code.
> 
> > Option 3 is updating the comments to describe the API implemented
> > rather than the one intended. This may be reasonable, given how few
> > users this library has.
> > Regarding DeepCopy,
> > > That the tests use the user-supplied instance is part of the API, too.
> > 
> > 
> > If a behavior is observable, that does not mean it must be treated like an API. (see also, [Hyrum's Law](https://www.hyrumslaw.com)) While it is good to avoid breaking users of this library, it is also good to avoid accumulating tech debt.
> 
> True. But this aspect of the package is useful, so it should be part of
> the API.

Are you referring to the nil panic fix, or the fact that test break if you DeepCopy a `Config`? I don't see how the latter is useful.

> 
> > Just as a side comment, the
> > [godoc](https://godoc.org/github.com/kubernetes-csi/csi-test/pkg/sanity)
> > for this package does not make it easy to figure out how to use this
> > package. Is there documentation elsewhere that users should be
> > reading?
> 
> Unfortunately not. New users probably discover it based on usage in
> other projects, which then also serves as example for how to use it.

It can always be improved in the future! May also be a `good-first-issue` task for a learning contributor.

pohly: > Are you referring to the nil panic fix, or the fact that test break if you DeepCopy a Config? I don't see how the latter is useful.

Shared ownership of Config with the ability to modify it later is useful in a Ginkgo suite due to the way how tests are run: there's registration of tests at init time, then code from the caller is executed (custom BeforeEach), then the code of each test runs. But that's a discussion for another time. For now let's simply ensure that code using 2.1.0 can update to 2.3.0 without having to change how the package is used (no API change).

@davidz627 can you lgtm to get this merged?

PR #224, opened 2019-10-04
Go modules

pohly:
**What type of PR is this?**
/kind cleanup

We want to use `go mod` instead of `dep` for dependency management.

```release-note
added Go module support and thus can be built also outside of GOPATH
```


pohly:
@msau42: this PR is ready for merging. It uses the latest `csi-lib-utils`, the one with the check that `go mod tidy` was used.

@bswartz: does this look okay? It replaces your PR #188.

While working on the blog article about `csi-sanity`, I noticed that `Go module` support has at least one tangible advantage: one no longer has to explain that the code must be checked out under `$GOPATH`. I think we can merge this and do a 2.3.0 release.

bswartz:
Thanks for doing this work

PR #225, opened 2019-10-06
sanity: check the error after removing MountTargetLocation

ZeroMagic:
**What type of PR is this?**
/kind bug

When developing a CSI driver, I used `csi-sanity` to test it. The test `should work` in `Node Service` worked well for the first time. But an error appeared for the second time. After reading the source code of sanity tests and debugging for a while, I found that I should delete the target directory `/tmp/csi-mount/target` in `NodeUnpublishVolume`. 
I always thought all subdirectories in `/tmp/csi-mount` would be deleted recursively by sanity test. Because there is no error to remind me. 
So this is the reason why I think we need to check the removing operation. With checking, it will help the developer to locate the error and improve development efficiency.

Fixes #

```release-note
check the error after removing MountTargetLocation
```

### DIFF
--- a/pkg/sanity/sanity.go
+++ b/pkg/sanity/sanity.go
@@ -291,8 +291,13 @@ func (sc *TestContext) Setup() {
 // allocated by Setup.
 func (sc *TestContext) Teardown() {
 	// Delete the created paths if any.
-	removeMountTargetLocation(sc.TargetPath, sc.Config.RemoveTargetPathCmd, sc.Config.RemoveTargetPath, sc.Config.RemovePathCmdTimeout)
-	removeMountTargetLocation(sc.StagingPath, sc.Config.RemoveStagingPathCmd, sc.Config.RemoveStagingPath, sc.Config.RemovePathCmdTimeout)
+	err1 := removeMountTargetLocation(sc.TargetPath, sc.Config.RemoveTargetPathCmd, sc.Config.RemoveTargetPath, sc.Config.RemovePathCmdTimeout)
+	err2 := removeMountTargetLocation(sc.StagingPath, sc.Config.RemoveStagingPathCmd, sc.Config.RemoveStagingPath, sc.Config.RemovePathCmdTimeout)
+	failures := InterceptGomegaFailures(func() {
+		Expect(err1).NotTo(HaveOccurred(), "failed to delete target directory %s", sc.TargetPath)
+		Expect(err2).NotTo(HaveOccurred(), "failed to delete staging directory %s", sc.StagingPath)
+	})
+	Expect(failures).To(BeEmpty())
 
 	// We intentionally do not close the connection to the CSI
 	// driver here because the large amount of connection attempts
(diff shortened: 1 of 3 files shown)


pkg/sanity/sanity.go
```diff
@@ -252,8 +252,10 @@ func (sc *SanityContext) Setup() {
 
 func (sc *SanityContext) Teardown() {
 	// Delete the created paths if any.
-	removeMountTargetLocation(sc.TargetPath, sc.Config.RemoveTargetPathCmd, sc.Config.RemoveTargetPath, sc.Config.RemovePathCmdTimeout)
-	removeMountTargetLocation(sc.StagingPath, sc.Config.RemoveStagingPathCmd, sc.Config.RemoveStagingPath, sc.Config.RemovePathCmdTimeout)
+	err := removeMountTargetLocation(sc.TargetPath, sc.Config.RemoveTargetPathCmd, sc.Config.RemoveTargetPath, sc.Config.RemovePathCmdTimeout)
+	Expect(err).NotTo(HaveOccurred(), "failed to delete target directory %s", sc.TargetPath)
```
pohly: If the check fails, `Teardown` will abort and skip the next `removeMountTargetLocation`. It would be nicer if it did both calls, then checked that both errors are nil.

ZeroMagic:
```go
err1 := removeMountTargetLocation(sc.TargetPath, sc.Config.RemoveTargetPathCmd, sc.Config.RemoveTargetPath, sc.Config.RemovePathCmdTimeout)
err2 := removeMountTargetLocation(sc.StagingPath, sc.Config.RemoveStagingPathCmd, sc.Config.RemoveStagingPath, sc.Config.RemovePathCmdTimeout)
Expect(err1).NotTo(HaveOccurred(), "failed to delete target directory %s", sc.TargetPath)
Expect(err2).NotTo(HaveOccurred(), "failed to delete staging directory %s", sc.StagingPath)
```
Do you mean that I need to modify it to be like this?

pohly: Yes, or have a single assertion that both errors are nil. What you have may be simpler.

ZeroMagic: done. Now it can show all failures.

pohly:
> It seems that mock driver has the same problem. The mock driver also does not delete the target directory.

Yes, the semantic around that directory has been confused in the past. Kubernetes itself still does it wrong and therefore the mock driver has a flag to make it work with Kubernetes as well as csi-test (https://github.com/kubernetes-csi/csi-test/commit/b634effe5dd9782fd31af9b2acfa0c50b3c5966d).

Can you look into fixing the delete code path?

ZeroMagic:
Ok, I will add another commit in the PR to fix it.

release-tools/prow.sh
```diff
@@ -890,7 +890,6 @@ EOF
 #!/bin/sh
 if ! kubectl exec "${CSI_PROW_SANITY_POD}" -c "${CSI_PROW_SANITY_CONTAINER}" -- rmdir "\$@"; then
     kubectl exec "${CSI_PROW_SANITY_POD}" -c "${CSI_PROW_SANITY_CONTAINER}" -- rm -rf "\$@"
-    exit 1
```
ZeroMagic: I have no idea about the reason for exiting. But I think we should not exit the script here because we do need to delete the directory. @pohly 

After removing it, all tests pass. Currently, the failure reason is that I should change it in `csi-release-tools` repo.

pohly: The `exit 1` is called if a simple `rmdir` has failed, typically because the driver hasn't remove its content from that directory. Why do you think that it needs to be removed? All that it does is cause the script and thus the test to fail; the cleanup still happens.

As you said, if this is a valid change, then it needs to go into `csi-release-tools` first.

ZeroMagic: In my opinion, whether `rmdir`  succussed or not, the directory will be removed in the original script. So the exit will not make sense.

Issue #223, opened 2019-10-02
Sanity: "pagination should detect volumes added between pages" removes volumes it didn't create & other issues


davidz627:
This test specifically checks for "existing volumes" by doing a `ListVolumes` call then afterwards does `removing all listed volumes`

This means that this test could removes all volumes listed even if the volume was not created by this test. This could cause data loss depending on how a consumer is using CSI sanity.

This brings up a larger problem I have with this test too - either
1) You are consuming csi-sanity and mocking out the back-end (this is what I do) -> this test is way too complicated and depends on the workings of pagination and doesn't feel like a simple "sanity test" like the rest of the package
2) You are not mocking out the backend -> this test has potential for data loss

msau42:
I agree that the test should only remove volumes created by the test.
/kind bug
/help

timoreimann:
I'd be happy to take a stab at this one.

pohly:
Note that the test is currently disabled (https://github.com/kubernetes-csi/csi-test/pull/226) because it was so problematic.

PR #226, opened 2019-10-10
sanity: Disable ListVolumes pagination test

avalluri:
**What type of PR is this?**
/kind bug

Volume created by testcase is not adding to cleanup list, hence it results in
volume leaking. This change fixes this issue.

The side effect of this testcase fix discovered a minor issue in mock driver implementation.
Which was hidden due to the leftover volume created by above testcase.

```release-note
NONE
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -259,11 +259,18 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *Sanity
 			Expect(len(vols.GetEntries())).To(Equal(totalVols))
 		})
 
-		It("pagination should detect volumes added between pages and accept tokens when the last volume from a page is deleted", func() {
+		// Disabling this below case as it is fragile and results are inconsistent
+		// when no of volumes are different. The test might fail on a driver
+		// which implements the pagination based on index just by altering
+		// minVolCount := 4 and maxEntries := 3
+		// Related discussion links:
+		//  https://github.com/intel/pmem-csi/pull/424#issuecomment-540499938
+		//  https://github.com/kubernetes-csi/csi-test/issues/223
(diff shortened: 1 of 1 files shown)


pohly:
As discussed in https://github.com/intel/pmem-csi/pull/424#issuecomment-540499938, the `ListVolume` pagination test is fragile. @avalluri showed there that it fails when the number of volumes is different. It probably also fails when csi-sanity isn't the only user of the cluster and other volumes are getting created or deleted concurrently.

I propose to disable the test in addition to fixing the volume leak. If someone has a good idea for improving the test, then they can do that later.

msau42:
Yes, @davidz627 opened up https://github.com/kubernetes-csi/csi-test/issues/223 outlining various problems with the ListVolumes tests.  I am ok with disabling the test because the issue in #223 I think is serious enough.

mock/service/controller.go
```diff
@@ -73,7 +73,7 @@ func (s *service) CreateVolume(
 		case *csi.VolumeContentSource_Volume:
 			vid := req.GetVolumeContentSource().GetVolume().GetVolumeId()
 			// Check if the source volume exists.
-			if volID, _ := s.findVolNoLock("id", vid); volID > 0 {
+			if volID, _ := s.findVolNoLock("id", vid); volID >= 0 {
```
pohly: This is an unrelated change. Can you move it to a separate PR?

avalluri: > @avalluri please disable "pagination should detect volumes added between pages and accept tokens when the last volume from a page is deleted" by replacing `It` with `XIt`, with a comment that links to #223 and [intel/pmem-csi#424 (comment)](https://github.com/intel/pmem-csi/pull/424#issuecomment-540499938).

Done.

avalluri: > This is an unrelated change. Can you move it to a separate PR?

Done: #227

PR #227, opened 2019-10-11
Fix mock driver

avalluri:
**What type of PR is this?**
/kind bug
/kind failing-test

`"should create volume from an existing source volume"`  test on mock driver fails when creating a new volume from existing volume stored at index 0. due to the wrong check in the driver code. This change fixes the issue.

```release-note
None
```


pohly:
/ok-to-test

Which test is failing?

Please remove the unrelated commit for the pagination tests.

avalluri:
> Which test is failing?

I hit with `"should create volume from an existing source volume" ` test failure as a side effect of #226.
> 
> Please remove the unrelated commit for the pagination tests.

Done, Thanks.

PR #229, opened 2019-10-18
add repetition loop to test idempotency

okartau:
Repeated volume operations helps to test the idempotency requirement.

**What type of PR is this?**
/kind feature

CSI spec requires driver to tolerate repeated volume operations by the idempotency requirements.
In [pmem-csi](https://github.com/intel/pmem-csi) driver context we discovered that some volume operations failed when requests were repeated. i.e. we failed to conform to idempotency requirements. However, the test suite does not repeat volume options in most cases, so here is attempt to add this.
There is config option that makes it possible to select repeat count, or bypass the repeated part.

If a component using csi-test does not implement all idempotency requirements, this addition may start to expose some new failures.

```release-note
Added (optional) repeated operations to test idempotency
```


okartau:
The failure in
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-test/229/pull-kubernetes-csi-csi-test/1185891170576764931
seems to point out the idempotency related issue in the hostpath driver, where repated NodeUnpublish does not succeed?
We had a similar one in pmem-csi when I started to add those repeated tests.

pohly:
> The failure in
https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-csi_csi-test/229/pull-kubernetes-csi-csi-test/1185891170576764931
seems to point out the idempotency related issue in the hostpath driver, where repated NodeUnpublish does not succeed?

In that case we'll have to fix the hostpath driver first such that it passes these new tests.

pohly:
> Please add a release note in the PR description. 

Still open.

okartau:
> > Please add a release note in the PR description.
> 
> Still open.

I added a short note, is this good?

pohly:
> repeated *Node* Volume operations

That note made me realized that the repetition count really is only applied to node operations - somehow I missed that earlier.

All operations must be idempotent, including `CreateVolume` and `ControllerPublishVolume`. Can you add that and update the changelog entry accordingly?

It would be nicer if the idempotency test wasn't under `node.go` because some CSI driver might only support the controller operations (for example, during initial development), but I think it's okay as it is because eventually the node operations are needed and then one can run the test.

pohly:
So this is going to keep failing until we update to testing against an updated hostpath driver, but otherwise looks good.

pkg/sanity/sanity.go
```diff
@@ -155,6 +155,16 @@ type TestConfig struct {
...
+
+	// Repeat count for Volume operations to test idempotency requirements.
+	// csi-test/pkg/sanity/node.go:runControllerTest() optionally runs
+	// added variant set of tests which repeats those Volume operations
+	// that are required to be idempotent, based on this count value.
+	// -1 : Skip idempotency test
+	//  0 : Use default value, 10 repetitions
+	//  n : repeat n times
+	// Note that missing explicit value leads to default repetition count.
+	IdempotentCount int
```
pohly: It just occurred to me that we no longer need this special logic with "0" meaning "default repetitions" because the default can now be set directly in https://github.com/kubernetes-csi/csi-test/blob/4560adb95d44a5ffeac46ea81fbb108979974e80/pkg/sanity/sanity.go#L188-L200

okartau:
rebased and pushed, to see fresh CI results after merge of https://github.com/kubernetes-csi/csi-driver-host-path/pull/139 which (I believe) should improve the testing result

okartau:
> should improve the testing result

but not yet as it was merged in master and tester uses fixed version.

pohly:
> So this is going to keep failing until we update to testing against an updated hostpath driver, but otherwise looks good.

https://github.com/kubernetes-csi/csi-test/pull/247 was merged, so we now should be testing with 1.3.0-rc4 of the hostpath driver, which in turn has the idempotency fix. Let's see...

/retest

pohly:
> #247 was merged, so we now should be testing with 1.3.0-rc4 of the hostpath driver, which in turn has the idempotency fix. Let's see...

Oh, bummer. Doesn't work like that because we deploy the hostpath driver which is specified in the kubernetes-1.17 deployment, and *that* still is 1.2.0 until we do a proper release of the hostpath driver.

I verified that the test passes with 1.3.0-rc4 by overriding the version in .prow.sh:

```
diff --git a/.prow.sh b/.prow.sh
index 81d4a1c..8c0e758 100755
--- a/.prow.sh
+++ b/.prow.sh
@@ -15,6 +15,10 @@
 # This repo supports and wants sanity testing.
 CSI_PROW_TESTS_SANITY=sanity
 
+# Temporarily test with the newer hostpath driver. Can be removed when
+# release-tools/prow.sh switches to CSI_PROW_DRIVER_VERSION v1.3.0.
+export HOSTPATHPLUGIN_TAG=v1.3.0-rc4
+
 . release-tools/prow.sh
 
 # Here we override "install_sanity" to use the pre-built one.
```

We could add that to this PR, but then we'll remove it again (hopefully) soon, so let's just keep this PR open for a bit longer.

okartau:
The base point was older, it still failed as was using 1.2.0 hostpath driver. 
Rebased and pushed.

pohly:
> The base point was older, it still failed as was using 1.2.0 hostpath driver.

Testing happens after merging, so no, that wasn't the reason. See my other comment.

pohly:
@okartau can you rebase, please?

okartau:
force-pushed after manual rebase.

pohly:
@okartau it's still failing. Can you have a look?

pohly:
> Oh, bummer. Doesn't work like that because we deploy the hostpath driver which is specified in the kubernetes-1.17 deployment, and that still is 1.2.0 until we do a proper release of the hostpath driver.

Just discovered that again... this is still a problem. PR #252 gets us to 1.3.0.

pohly:
> PR #252 gets us to 1.3.0.

That one was merged.

/retest

pkg/sanity/sanity.go
pohly: @okartau: this is still open.

okartau:
pushed new version, is this what you meant?
I also discovered the other changed part, Controller test, has to be prepared for count value being less than 1, as there we do not have different calling but we use simple loop. So, to have testing without repeat, we have to force loop count as one. Is this good approach, or should we add code to explicitly ha another  calling branch if idempotency repeater is activated?

pkg/sanity/sanity.go
```diff
@@ -156,6 +156,14 @@ type TestConfig struct {
 	// generator for valid Volume and Node IDs. Defaults to
 	// DefaultIDGenerator.
 	IDGen IDGenerator
+
+	// Repeat count for Volume operations to test idempotency requirements.
+	// some tests can optionally run repeated variants for those Volume operations
+	// that are required to be idempotent, based on this count value.
+	// -1 : Skip idempotency test
+	//  n : repeat n times
+	// Note that missing explicit value leads to default repetition count.
```
pohly: There isn't really a "missing explicit value" anymore. You can replace that last sentence with "NewTestConfig() by default enables idempotency testing."

pkg/sanity/node.go
```diff
@@ -618,167 +795,18 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
...
-				context.Background(),
-				&csi.ControllerUnpublishVolumeRequest{
-					VolumeId: vol.GetVolume().GetVolumeId(),
-					NodeId:   ni.GetNodeId(),
-					Secrets:  sc.Secrets.ControllerUnpublishVolumeSecret,
-				},
-			)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(controllerunpubvol).NotTo(BeNil())
+		if sc.Config.IdempotentCount < 0 {
```
pohly: `<= 0` ?

pkg/sanity/controller.go
```diff
@@ -1110,19 +1117,21 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
...
-				&csi.ControllerPublishVolumeRequest{
-					VolumeId:         vol.GetVolume().GetVolumeId(),
-					NodeId:           ni.GetNodeId(),
-					VolumeCapability: TestVolumeCapabilityWithAccessType(sc, csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER),
-					Readonly:         false,
-					Secrets:          sc.Secrets.ControllerPublishVolumeSecret,
-				},
-			)
-			Expect(err).NotTo(HaveOccurred())
+			for i := 0; i < idempotentcount; i++ {
```
pohly: This modifies the existing "ControllerPublishVolume should return appropriate values (no optional values added)" test. Idempotency of the other controller functions isn't covered.

Can you use the same pattern as for the node testing where a helper function runs the full test such that *every* function is repeated n times? The simple version of the test then only does it once.

okartau: > same pattern as for the node

good point, that is a cleaner way.
While looking for tests that can be repeated, I notice that we have two code blocks
that are functionally identical, only differing by few strings (volume name and By), and minor ordering of statements (not meaningful functionally).
One block is lines 1068..1152  (master branch)
another is 1385..1470
The idea (by names) seems to be that first one tests Publish and 2nd one Unpublish,
but as both perform Create-Publish-Unpublish-Delete, these both test same thing.
If we really want to have such repeated test group, it makes sense to turn this code into function
(which I BTW do anyway in repetition-enabling change).
But we can also ask question, can we drop the 2nd instance, as it does not add anything new,
as both Publish and Unpublish do get tested same way already in 1st instance.

pohly:
> can we drop the 2nd instance, as it does not add anything new

Yes, we can.

okartau:
and it's ok to do it here even if it's opposite of this PR topic? (and without creating an issue).
I will add a separate commit then.

pohly:
> and it's ok to do it here even if it's opposite of this PR topic?

Yes. It's part of refactoring the code to support a new test.

okartau:
Pushed version with Volume lifecycle separated as function which can be called with repeat-count argument. Also, removing the duplicate part from Unpublish side, in separate commit.
I think we can improve `It` and `By strings a bit, currently used 
`It("should return appropriate values (no optional values added)"`
(which I did not changed yet in this PR)
seems to be used as copied around template used in many places and not a good fit,
as we are not checking any optional values in that code.

pkg/sanity/controller.go
```diff
@@ -1065,90 +1072,14 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
...
-				&csi.DeleteVolumeRequest{
-					VolumeId: vol.GetVolume().GetVolumeId(),
-					Secrets:  sc.Secrets.DeleteVolumeSecret,
-				},
-			)
-			Expect(err).NotTo(HaveOccurred())
-			cl.UnregisterVolume(name)
+		It("should return appropriate values (no optional values added), with idempotency", func() {
+			By("lifecycle repeated")
+			VolumeLifecycle(n, c, sc, cl, idempotentcount)
```
pohly: You are right, "should return appropriate values (no optional values added)" doesn't make a lot of sense. I would go one step further than merely renaming it: this and the idempotency variant are basically a full test for "Controller Service", not just for "Controller Service ControllerPublishVolume", which is where it is currently defined.

Can you move it out of `Describe("ControllerPublishVolume", func() {` and instead use `Describe("works", ...` for the simple test and `Describe("is idempotent", ...` for the one with repetition?

pohly: Updated the comment: it now uses exactly the same description as the corresponding node tests.

okartau: pushed new version where I tried to follow these ideas.

okartau:
This lacked skip check so it failed the CI run. Added skip andtry pushed.
Q: Is the location of those what I called "Lifecycle" tests now good, it's between Publish and Unpublish tests. Or would it be better to put it after Unpublish part.

pkg/sanity/controller.go
```diff
@@ -1360,6 +1275,22 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 		})
 	})
 
+	Describe("ControllerVolumeLifecycle", func() {
```
pohly: "ControllerVolumeLifecycle" is plain English, not the name of a gRPC call as in the other `Describe` cases. Better just use "volume lifecycle". "Controller" can be left out because the top-level `Describe` already is "Controller Service [Controller Server]" - no need to repeat it here.

pkg/sanity/controller.go
```diff
@@ -123,6 +123,7 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 	AfterEach(func() {
 		cl.DeleteVolumes()
 	})
+	idempotentcount := sc.Config.IdempotentCount
```
pohly: This local variable is unnecessary. Better drop it.

okartau:
addressed latest comments and force-pushed.

pohly:
Let's see if someone else has an opinion about this. If not, then I'll approve.

PR #232, opened 2019-10-28
Fix go module requirements for semantic versioning

### LINKED ISSUE #231: Go mod breaks importing
Since Go 1.13 (I think) it is required that a `v2` release or higher contain the major version in the module directive (https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher).

Since this is not set, it effectively breaks updating to `v2.3.0`:

```
go get github.com/kubernetes-csi/csi-test@v2.3.0
...
require github.com/kubernetes-csi/csi-test: version "v2.3.0" invalid: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

The solution is rather simple, but it may be a breaking change for some users.

Ntr0:
As stated in this post:
https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher
modules that released a version higher than 1 need to put the version
into the package path.

**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

When 2.3.0 introduced go modules, the package paths where not updated. Therefore
this module cannot be imported as version 2.3.0 since go mod will complain: 
```console
go.mod:12: require github.com/kubernetes-csi/csi-test: version "v2.3.0" invalid: module contains a go.mod file, so major version must be compatible: should be v0 or v1, not v2
```

Fixes #231 

```release-note
Update package path to v3. Vendoring with dep depends on https://github.com/golang/dep/pull/1963 or the workaround described in v3/README.md.
```

### DIFF
--- a/cmd/csi-sanity/sanity_test.go
+++ b/cmd/csi-sanity/sanity_test.go
@@ -21,7 +21,7 @@ import (
 	"os"
 	"testing"
 
-	"github.com/kubernetes-csi/csi-test/pkg/sanity"
+	"github.com/kubernetes-csi/csi-test/v2/pkg/sanity"
 )
 
 const (
--- a/cmd/csi-sanity/sanity_test.go
+++ b/cmd/csi-sanity/sanity_test.go
@@ -21,7 +21,7 @@ import (
 	"os"
 	"testing"
 
-	"github.com/kubernetes-csi/csi-test/v2/pkg/sanity"
+	"github.com/kubernetes-csi/csi-test/v3/pkg/sanity"
 )
 
 const (
--- a/cmd/mock-driver/main.go
+++ b/cmd/mock-driver/main.go
@@ -24,8 +24,8 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/kubernetes-csi/csi-test/driver"
-	"github.com/kubernetes-csi/csi-test/mock/service"
+	"github.com/kubernetes-csi/csi-test/v2/driver"
+	"github.com/kubernetes-csi/csi-test/v2/mock/service"
 )
 
 func main() {
--- a/cmd/mock-driver/main.go
+++ b/cmd/mock-driver/main.go
@@ -24,8 +24,8 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/kubernetes-csi/csi-test/v2/driver"
-	"github.com/kubernetes-csi/csi-test/v2/mock/service"
+	"github.com/kubernetes-csi/csi-test/v3/driver"
+	"github.com/kubernetes-csi/csi-test/v3/mock/service"
 )
 
 func main() {
--- a/driver/mock.go
+++ b/driver/mock.go
@@ -19,7 +19,7 @@ package driver
 import (
 	"net"
 
-	"github.com/kubernetes-csi/csi-test/v2/utils"
+	"github.com/kubernetes-csi/csi-test/v3/utils"
 	"google.golang.org/grpc"
 )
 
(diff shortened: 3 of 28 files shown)


Ntr0:
@pohly It looks like it works. This approach seems reasonable to me, as I won't expect csi-test to maintain different versions of the package. So keeping the semantic version in the main go.mod and having these symlinks looks good to me.

msau42:
> it is so intrusive for the git history

Doesn't git mv retain history?

pohly:
Michelle Au <notifications@github.com> writes:
>> it is so intrusive for the git history
>
> Doesn't git mv retain history?

"git mv" is the same as deleting a file and adding it under a different
name. It doesn't have any special support for keeping history. This is
all handled when analyzing the commits: some operations detect such
renames and follow them, some don't.

For example, "git log v3" would stop at the commit that introduced the
directory. "git log v3/pkg/sanity/sanity.go" would work.

pohly:
> Seems fine to me if the file still retains history.

There are other drawbacks, like having to maintain the `go.mod` file also in the `v3` directory. I don't know what it means for vendoring. Probably breaks our tooling and/or requires special casing.

> I think it's kind of messy and confusing to have symlinks

It's an interim solution until either "dep" gets enhanced (PR is pending) or we can convince ourselves that supporting "dep" isn't necessary anymore. I prefer symlinks over messing with the directory structure.

pohly:
If there's no consensus on how to support "dep", then I propose that we simply don't do anything special to support it, i.e. remove 0163355 from this PR again.

Then "dep" users either have to switch to "go mod" or use a patched dep version (https://github.com/golang/dep/pull/1963).

pohly:
> Can you also add a release note indicating the potential future version of dep that will be needed?

It seems uncertain which version of dep will have https://github.com/golang/dep/pull/1963 included; might be a patch or a major release. Better link to the PR.

Ntr0:
@pohly Updated Release Notes to point out the breaking change for dep users

pohly:
> Updated Release Notes to point out the breaking change for dep users

Sorry for the back and forth, but when we discussed this yesterday in the Kubernetes-CSI standup meeting, @msau42 said that we should continue to support importing with normal dep. We settled on the approach with symlinks as the lesser evil.

Can you remove the revert commit from your branch and augment my commit with the following `v3/README.md`?

```
This directory mirrors the source code via symlinks.
This makes it possible to vendor v3.x releases of
csi-test with `dep` versions that do not support
semantic imports. Support for that is currently
[pending in dep](https://github.com/golang/dep/pull/1963).

If users of dep have enabled pruning, they must disable if
for csi-test in their Gopk.toml, like this:

    [prune]
      go-tests = true
      unused-packages = true

      [[prune.project]]
        name = "github.com/kubernetes-csi/csi-test"
        unused-packages = false
```

Can you also adapt the release note accordingly?

Thanks!

pohly:
The double revert looks a bit weird, personally I would use `git rebase -i` and just drop both commits. Other than that this looks good. I'll leave the final decision to @msau42 .

pohly:
Wait, the release note still needs to be updated:

```
Update package path to v3. Vendoring with dep is not supported anymore, at least until https://github.com/golang/dep/pull/1963 has been merged.
```
=>
```
Update package path to v3. Vendoring with dep depends on https://github.com/golang/dep/pull/1963 or the workaround described in v3/README.md.
```

pohly:
/lgtm

There's a minor spelling nit in the README.md (my bad): "must enable *it*", but I propose to merge as-is to get this done.

Leaving the approval to @msau42.

/assign @msau42

PR #236, opened 2019-10-31
Add snapshot parameter flag

### LINKED ISSUE #235: Add snapshot parameters file flag
When creating snapshot,  VolumeSnapshotClass's parameters will passed through CreateSnapshotRequest.parameters. We could add a flag to pass VolumeSnapshotClass parameters
 in sanity test just like StorageClass parameters.

wnxn:
Signed-off-by: Xin Wang <wileywang@yunify.com>

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

Fixes #235 

```release-note
Add '--csi.testsnapshotparameters' flag
```

### DIFF
--- a/pkg/sanity/sanity.go
+++ b/pkg/sanity/sanity.go
@@ -76,6 +76,10 @@ type Config struct {
 
 	JUnitFile string
 
+	// TestSnapshotParametersFile for setting CreateVolumeRequest.Parameters.
+	TestSnapshotParametersFile string
+	TestSnapshotParameters     map[string]string
+
 	// Callback functions to customize the creation of target and staging
 	// directories. Returns the new paths for mount and staging.
 	// If not defined, directories are created in the default way at TargetPath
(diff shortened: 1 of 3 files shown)


pkg/sanity/sanity.go
```diff
@@ -163,18 +165,30 @@ func newContext(reqConfig *Config) *SanityContext {
...
+			panic(fmt.Sprintf("failed to read file %q: %v", volParamPath, err))
 		}
 		err = yaml.Unmarshal(yamlFile, &reqConfig.TestVolumeParameters)
 		if err != nil {
 			panic(fmt.Sprintf("error unmarshaling yaml: %v", err))
 		}
 	}
 
+	snapParamPath := reqConfig.TestSnapshotParametersFile
+	if len(snapParamPath) != 0 {
```
pohly: Instead copying the unmarshaling code from the `reqConfig.TestVolumeParametersFile` handling above, can you move it into a helper function?

pkg/sanity/sanity.go
```diff
@@ -69,10 +69,12 @@ type Config struct {
...
 
 	// Target size for ExpandVolume requests. If not specified it defaults to TestVolumeSize + 1 GB
-	TestVolumeExpandSize      int64
-	TestVolumeParametersFile  string
-	TestVolumeParameters      map[string]string
-	TestNodeVolumeAttachLimit bool
+	TestVolumeExpandSize       int64
+	TestVolumeParametersFile   string
+	TestVolumeParameters       map[string]string
+	TestSnapshotParametersFile string
```
pohly: Can you add the new fields in a separate block (i.e. don't mix with `TestVolumeExpand*`) and add documentation? I know it's missing also for the other fields, but that's not a good example to follow :man_shrugging:

pkg/sanity/sanity.go
wnxn: Thanks for your review. I will amend my commit.

pkg/sanity/sanity.go
```diff
@@ -357,6 +354,22 @@ func loadSecrets(path string) (*CSISecrets, error) {
 	return &creds, nil
 }
 
+// loadMap reads a map from given file path.
+func loadMap(path string) map[string]string {
```
pohly: I think you can make this more generic:
```
func loadFromFile(from string, to interface{}) {
    if len(path) != 0 {
...
        err = yaml.Unmarshal(yamlFile, to)
    }
}
```

Then call it with:
```
loadFromFile(reqConfig.TestSnapshotParametersFile, &reqConfig.TestSnapshotParameters)
```

Untested, but I think it'll work, and then we have something that could also become useful for populating other data structures.

wnxn: OK, I will update and test it.

wnxn: Hi @pohly , The method is worked and I updated this PR. Would you please review again?

PR #242, opened 2019-12-16
Add NodeUnpublishVolume test for when the volume is missing

timoreimann:
**What type of PR is this?**
/kind feature

Adds `NodeUnpublishVolume` test for when the volume is missing.

The spec mandates that such cases should return NOT_FOUND responses ([reference](https://github.com/container-storage-interface/spec/blob/4731db0e0bc53238b93850f43ab05d9355df0fd9/spec.md#nodeunpublishvolume-errors)).

The mock driver already implements the right behavior.

```release-note
Add NodeUnpublishVolume test for when the volume is missing
```

### DIFF
--- a/pkg/sanity/node.go
+++ b/pkg/sanity/node.go
@@ -241,6 +241,21 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
 			Expect(ok).To(BeTrue())
 			Expect(serverError.Code()).To(Equal(codes.InvalidArgument))
 		})
+
+		It("should fail when the volume is missing", func() {
+
+			_, err := c.NodeUnpublishVolume(
+				context.Background(),
+				&csi.NodeUnpublishVolumeRequest{
+					VolumeId:   sc.Config.IDGen.GenerateUniqueValidVolumeID(),
+					TargetPath: sc.StagingPath,
+				})
+			Expect(err).To(HaveOccurred())
+
+			serverError, ok := status.FromError(err)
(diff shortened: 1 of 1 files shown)


pkg/sanity/node.go
```diff
@@ -241,6 +241,21 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
...
+
+		It("should fail when the volume is missing", func() {
+
+			_, err := c.NodeUnpublishVolume(
+				context.Background(),
+				&csi.NodeUnpublishVolumeRequest{
+					VolumeId:   sc.Config.IDGen.GenerateUniqueValidVolumeID(),
+					TargetPath: sc.StagingPath,
+				})
+			Expect(err).To(HaveOccurred())
```
pohly: `Expect.To` without some additional description leads to failure messages that cannot be understood without locating the source code. I know, most of the existing assertions are like that, but for new ones let's use `Expect.To(..., "some description")`.

pkg/sanity/node.go
```diff
@@ -241,6 +241,21 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
...
+
+			serverError, ok := status.FromError(err)
+			Expect(ok).To(BeTrue())
```
pohly: Same here.

timoreimann:
@pohly sorry, I haven't had a chance to get back to this PR in time before it got merged. I'll address your comments in a follow-up PR.

PR #243, opened 2019-12-16
Update pkg/sanity Go driver usage README instructions

timoreimann:
**What type of PR is this?**
/kind documentation

The change updates the `pkg/sanity` Go driver usage instructions in the README.

It seems like the approach to putting together a test configuration has changed since it was previously written down.

```release-note
NONE
```

/sig storage


timoreimann:
/assign @pohly 

since you seem to be on a doc updating roll right now. :)

PR #249, opened 2020-02-13
Mock topology testing

pohly:
**What type of PR is this?**
/kind failing-test

For testing the upcoming "unset selected node" feature in external-provisioner.

Related-to: https://github.com/kubernetes/kubernetes/issues/72031

This could be merged as-is, but it might also make sense to first review the other PRs.

```release-note
NONE
```

### DIFF
--- a/cmd/mock-driver/main.go
+++ b/cmd/mock-driver/main.go
@@ -37,6 +37,7 @@ func main() {
 	flag.StringVar(&config.DriverName, "name", service.Name, "CSI driver name.")
 	flag.Int64Var(&config.AttachLimit, "attach-limit", 2, "number of attachable volumes on a node")
 	flag.BoolVar(&config.NodeExpansionRequired, "node-expand-required", false, "Enables NodeServiceCapability_RPC_EXPAND_VOLUME capacity.")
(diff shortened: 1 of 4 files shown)


mock/service/controller.go
```diff
@@ -58,6 +63,18 @@ func (s *service) CreateVolume(
 		return nil, status.Errorf(codes.OutOfRange, "Requested capacity %d exceeds maximum allowed %d", capacity, MaxStorageCapacity)
 	}
 
+	// Simulate "out of storage space" error?
+	if capacityLimit := req.GetParameters()["capacityLimit"]; capacityLimit != "" {
```
msau42: @jsafrane how soon is that error injection coming?

pohly: /hold

There's now a WIP PR for that in PR #251. Using that approach makes more sense than adding special cases like this one here to the code.

pohly:
I've stripped out the error injection and rewrote https://github.com/kubernetes/kubernetes/pull/88114 to use JavaScript error injection.

What's left is just the optional topology support. Can we merge this now? Then I'll do a csi-test release.

pohly:
That should also unblock https://github.com/kubernetes/kubernetes/pull/89041.

mock/service/controller.go
pohly: /hold cancel

PR #251, opened 2020-02-24
Add JavaScript hooks

tsmetana:
**What type of PR is this?**
/kind feature

In order to be able to test also various failure situations I've added a JavaScript interpreter and hooks to the CSI methods. The hooks content is meant to be configured by a ConfigMap containing JavaScript snippets. Furthermore if the snippet evaluates to an integer value it gets interpreted as a gRPC code to be immediately returned from the particular method. This allows for things like "fail twice with DeadlineExceeded then once with Aborted and then succeed" or even more complicated logic.

```release-note
The CSI mock driver behaviour can be tweaked with JavaScript hooks in the CSI calls.
```


tsmetana:
Thanks fro the comment -- it's what I was hoping for. :)

The idea comes from kitchen chats with @jsafrane and @bertinatto and that's about all the discussions I lead about it so far... Actually getting some opinions whether this makes sense and what exactly do we need is a main point of this PR.

Here's an usage example for illustration: Prepare a YAML file defining some hooks. Let's call it `hooks.yaml`:
```
Globals: |
  count = 0;
  console.log("Globals loaded, count set to " + count);
CreateVolumeStart: |
  count = count + 1;
  console.log("CreateVolumeStart: The value of count is " + count);
  if (count > 2) { __grpcCode.OK; } else { ` };
CreateVolumeEnd: |
  console.log("CreateVolumeEnd");
```
I turn it to a `ConfigMap`:
```
kubectl create configmap hooks-config --from-file hooks.yaml
```
Make my CSI mock driver to use it -- this is the sidecar definition YAMLs:
```
...
containers:
  ...
  - name: csi-mock-plugin
    ...
    args:
      - --hooks-file=/var/lib/hooks/hooks.yaml
    ...
    volumeMounts:
      ...
      - name: hooks-volume
        mountPath: /var/lib/hooks
...
volumes:
  - name: hooks-volume
    configMap:
      name: hooks-config
...
```
Now the driver finds the hooks definitions file and creates a JavaScript VM -- just one instance of it that exists through the life of the driver that will execute the JavaScript snippets at the defined places in the code. One special case is the `Globals` hook that is executed right after loading the file and it's intended for variable initializations, common function definitions, etc. I also hardcoded injection of the gRPC code constants for convenience.

Each code snippet is executed at given place in the code: in our example `CreateVolumeStart` at the beginning of the `CreateVolume` method in the driver. The result of the hook snipped is evaluated: it would be the value of the last expression executed. If the result is an integer value it gets interpreted as gRPC code and in case it's non-zero (not `code.OK`), the CSI driver method would return this as its response error.

In our example: I initialized the variable `count` in the `Globals` hook to zero. Each time we get into the `CreateVolume` method in the driver the hook is executed so `count` is increased by one (it's always the same variable) and for the first two calls the last expression of the code snipped is `__grpcCode.DEADLINEEXCEEDED` which causes the `CreateVolume` method to fail with the given code. Subsequent calls would return `0` so things would work normally. Also -- the `console.out()` allows me to log to stdout so I can see the messages from the hooks among the output from the driver itself...

Now, is this useful? Could I make it useful? I understand that passing one number in one direction is quite limited ability but I also don't want to make the code too complicated. It really depends on use-cases we might have.

pohly:
> Now, is this useful?

I find it useful already with the current limitations. My only concern is the amount of additional dependencies for the JS interpreter, but that cannot be avoided and IMHO isn't too bad.

Let's hear from others... @jsafrane, @msau42?

hooks-howto.md
```diff
@@ -0,0 +1,72 @@
...
+sidecar definition YAMLs:
+
+```
+...
+containers:
+  ...
+  - name: csi-mock-plugin
+    ...
+    args:
+      - --hooks-file=/var/lib/hooks/hooks.yaml
```
jsafrane: nit: `/etc/hooks/hooks.yaml`?

hooks-howto.md
```diff
@@ -0,0 +1,72 @@
...
+      name: hooks-config
+...
+```
+
+Now the driver finds the hooks definitions file and creates a JavaScript VM.
+Just one instance of the VM exists through the life of the driver. It executes
+the pieces of code as per the YAML file. One special case is the `Globals` hook
+that is executed right after loading the file and it's intended for variable
+initialization, common function definitions, etc. There are also pre-defined
+gRPC constants in the `__grpcCode` variable.
```
jsafrane: The constants should be listed. Or even better, linked.

hooks-howto.md
```diff
@@ -0,0 +1,72 @@
...
+
+Each code snippet is executed at given place in the code: in this example
+`CreateVolumeStart` at the beginning of the `CreateVolume` method in the driver.
+The result of the hook snipped is evaluated: it would be the value of the last
+expression executed. If the result is an integer value it gets interpreted as
+gRPC code and in case it's non-zero (not `code.OK`), the CSI driver method would
```
jsafrane: `__grpcCode.OK`

hooks-howto.md
```diff
@@ -0,0 +1,72 @@
...
+return this as its response error.
+
+In the example: The variable `count` was initialized to zero in the `Globals`
+hook. Each time the `CreateVolume` method is called the driver the hook is
```
jsafrane: "Each time the `CreateVolume` method is called the hook is executed"

mock/service/service.go
```diff
@@ -73,6 +119,33 @@ func New(config Config) Service {
...
+			RESOURCEEXHAUSTED: 8,
+			FAILEDPRECONDITION: 9,
+			ABORTED: 10,
+			OUTOFRANGE: 11,
+			UNIMPLEMENTED: 12,
+			INTERNAL: 13,
+			UNAVAILABLE: 14,
+			DATALOSS: 15,
+			UNAUTHENTICATED: 16
+		};`)
```
jsafrane: It would be useful to have this in some `globals.js` file in this repo, so it can be linked from documentation easily && edited without go knowledge.

The file itself can be either embedded into csi-mock binary via bindata or it can be shipped in the container with the driver.

mock/service/service.go
```diff
@@ -73,6 +119,33 @@ func New(config Config) Service {
...
+
+		_, err := s.hooksVm.Run(s.config.ExecHooks.Globals)
+		if err != nil {
+			fmt.Printf("Warning: Error encountered in the global exec hook: %v\n", err)
```
jsafrane: I would prefer `os.Exit`, otherwise the driver may do something unexpected in its CSI calls.

mock/service/service.go
```diff
@@ -185,3 +258,25 @@ func (s *service) getAttachCount(devPathKey string) int64 {
...
 }
+
+func (s *service) execHook(hookName string) (codes.Code, string) {
+	if s.hooksVm != nil {
+		script := reflect.ValueOf(*s.config.ExecHooks).FieldByName(hookName).String()
+		if len(script) > 0 {
+			result, err := s.hooksVm.Run(script)
+			if err != nil {
+				fmt.Printf("Exec hook %s error -- hook ignored: %v\n", hookName, err)
+				return codes.OK, ""
```
jsafrane: This fails on syntax errors and similar stuff. `os.Exit`?

jsafrane:
Regarding javascript interpreter dependencies. Yes, they're huge and it does not look that the interpreter is well maintained. On the other hand, it is used only for testing and any bugs there (or even CVEs) should not hurt anyone.

mock/service/service.go
jsafrane: Or, create `hook_globals.go` with a single string constant and link it from docs...

tsmetana: Also -- let's remove the `__grpc` object and leave just the constants there. This behaves unexpectedly when one makes a typo (no error encountered).

hooks-howto.md
tsmetana: `OK` now.

hooks-howto.md
tsmetana: Fixed.

mock/service/service.go
tsmetana: The new commid adds os.Exit(1) on error.

hooks-howto.md
tsmetana: Changed.

mock/service/service.go
tsmetana: Done.

mock/service/hooks-const.go
```diff
@@ -0,0 +1,24 @@
+package service
+
+// Predefinded constants for the JavaScript hooks, they must correspond to the
+// error codes used by gRPC, see:
+// https://github.com/grpc/grpc-go/blob/master/codes/codes.go
+const (
+	grpcJsCodes string = `OK = 0;
```
jsafrane: nit: `grpcJSCodes`

mock/service/service.go
```diff
@@ -26,6 +31,46 @@ var Manifest = map[string]string{
...
+	NodePublishVolumeStart         string `yaml:"NodePublishVolumeStart"`
+	NodePublishVolumeEnd           string `yaml:"NodePublishVolumeEnd"`
+	NodeUnpublishVolumeStart       string `yaml:"NodeUnpublishVolumeStart"`
+	NodeUnpublishVolumeEnd         string `yaml:"NodeUnpublishVolumeEnd"`
+	NodeExpandVolumeStart          string `yaml:"NodeExpandVolumeStart"`
+	NodeExpandVolumeEnd            string `yaml:"NodeExpandVolumeEnd"`
+	NodeGetCapabilities            string `yaml:"NodeGetCapabilities"`
+	NodeGetInfo                    string `yaml:"NodeGetInfo"`
+	NodeGetVolumeStatsStart        string `yaml:"NodeGetVolumeStatsStart"`
+	NodeGetVolumeStatsEnd          string `yaml:"NodeGetVolumeStatsEnd"`
```
jsafrane: Traditionally, json and yaml identifiers start with lowercase (`globals`, `nodeGetVolumeStatsEnd`, ...)

tsmetana: I meant for these to resemble the names of the methods... But... OK, I'll change it.

mock/service/hooks-const.go
tsmetana: Fixed.

PR #254, opened 2020-03-20
Log full error code

jsafrane:
Dump whole error into gRPCCall logs, so we can check error code in e2e tests.

**What type of PR is this?**
/kind feature

```release-note
"gRPCCall" logs on standard output now include "fullError" with complete error structure. Typically, it contains "code" and "message" with gRPC error.
```

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -219,15 +219,22 @@ func logGRPC(method string, request, reply interface{}, err error) {
 		Method   string
 		Request  interface{}
 		Response interface{}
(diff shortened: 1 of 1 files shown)


jsafrane:
This is necessary for checking if appropriate CSI method was called and failed with expected error code in new e2e tests in https://github.com/kubernetes/kubernetes/pull/89041

jsafrane:
@pohly @msau42, PTAL

PR #259, opened 2020-04-17
Improve ListSnapshot filtering tests

timoreimann:
**What type of PR is this?**
> /kind feature

We provide tests on `ListSnapshots` for filtering by snapshot IDs and source volume IDs. However, the tests do not cover the case were snapshot listing functionality exists in general but is missing
filtering capabilities: for this scenario the existing tests yield false positives because the fixtures set up a single snapshot only.

We improve the tests by creating a number of snapshots of which a subset is expected to be found.

```release-note
NONE
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -1340,35 +1340,71 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *TestContext
 	})
 
 	It("should return snapshots that match the specified snapshot id", func() {
+		// The test creates three snapshots: one that we intend to find by
+		// snapshot ID, and two unrelated ones that must not be returned by
+		// ListSnapshots.
 
-		By("creating a volume")
-		volReq := MakeCreateVolumeReq(sc, "listSnapshots-volume-1")
-		volume, err := c.CreateVolume(context.Background(), volReq)
+		By("creating first unrelated snapshot")
+		// Create volume source and afterwards the first unrelated snapshot.
+		volReq := MakeCreateVolumeReq(sc, "listSnapshots-volume-unrelated1")
+		volumeUnrelated1, err := c.CreateVolume(context.Background(), volReq)
 		Expect(err).NotTo(HaveOccurred())
 
-		By("creating a snapshot")
-		snapshotReq := MakeCreateSnapshotReq(sc, "listSnapshots-snapshot-1", volume.GetVolume().GetVolumeId())
-		snapshot, err := c.CreateSnapshot(context.Background(), snapshotReq)
+		snapshotReq := MakeCreateSnapshotReq(sc, "listSnapshots-snapshot-unrelated1", volumeUnrelated1.GetVolume().GetVolumeId())
+		snapshotUnrelated1, err := c.CreateSnapshot(context.Background(), snapshotReq)
 		Expect(err).NotTo(HaveOccurred())
 
+		By("creating target snapshot")
+		// Create volume source and afterwards the target snapshot.
+		volReq = MakeCreateVolumeReq(sc, "listSnapshots-volume-target")
+		volumeTarget, err := c.CreateVolume(context.Background(), volReq)
+		Expect(err).NotTo(HaveOccurred())
+
+		snapshotReq = MakeCreateSnapshotReq(sc, "listSnapshots-snapshot-target", volumeTarget.GetVolume().GetVolumeId())
+		snapshotTarget, err := c.CreateSnapshot(context.Background(), snapshotReq)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("creating second unrelated snapshot")
+		// Create volume source and afterwards the second unrelated snapshot.
+		volReq = MakeCreateVolumeReq(sc, "listSnapshots-volume-unrelated2")
+		volumeUnrelated2, err := c.CreateVolume(context.Background(), volReq)
+		Expect(err).NotTo(HaveOccurred())
+
+		snapshotReq = MakeCreateSnapshotReq(sc, "listSnapshots-snapshot-unrelated2", volumeUnrelated2.GetVolume().GetVolumeId())
+		snapshotUnrelated2, err := c.CreateSnapshot(context.Background(), snapshotReq)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("listing snapshots")
 		snapshots, err := c.ListSnapshots(
 			context.Background(),
-			&csi.ListSnapshotsRequest{SnapshotId: snapshot.GetSnapshot().GetSnapshotId()})
+			&csi.ListSnapshotsRequest{SnapshotId: snapshotTarget.GetSnapshot().GetSnapshotId()})
 		Expect(err).NotTo(HaveOccurred())
 		Expect(snapshots).NotTo(BeNil())
 		Expect(len(snapshots.GetEntries())).To(BeNumerically("==", 1))
 		verifySnapshotInfo(snapshots.GetEntries()[0].GetSnapshot())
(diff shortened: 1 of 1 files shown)


pkg/sanity/controller.go
```diff
@@ -1341,34 +1341,63 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *TestContext
 
 	It("should return snapshots that match the specified snapshot id", func() {
 
-		By("creating a volume")
-		volReq := MakeCreateVolumeReq(sc, "listSnapshots-volume-1")
-		volume, err := c.CreateVolume(context.Background(), volReq)
+		By("creating volumes")
+		volReq := MakeCreateVolumeReq(sc, "listSnapshots-volume-unrelated1")
```
xing-yang, line 1345: What does "unrelated" mean?  Please add a comment.

pkg/sanity/controller.go
```diff
@@ -1341,34 +1341,63 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *TestContext
...
+		volumeUnrelated1, err := c.CreateVolume(context.Background(), volReq)
 		Expect(err).NotTo(HaveOccurred())
 
-		By("creating a snapshot")
-		snapshotReq := MakeCreateSnapshotReq(sc, "listSnapshots-snapshot-1", volume.GetVolume().GetVolumeId())
-		snapshot, err := c.CreateSnapshot(context.Background(), snapshotReq)
+		volReq = MakeCreateVolumeReq(sc, "listSnapshots-volume-target")
```
xing-yang, line 1349: Add a comment

pkg/sanity/controller.go
```diff
@@ -1341,34 +1341,63 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *TestContext
...
+		volumeTarget, err := c.CreateVolume(context.Background(), volReq)
+		Expect(err).NotTo(HaveOccurred())
+
+		volReq = MakeCreateVolumeReq(sc, "listSnapshots-volume-unrelated2")
```
xing-yang, line 1353: Add comment

pkg/sanity/controller.go
```diff
@@ -1341,34 +1341,63 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *TestContext
...
+		volumeUnrelated2, err := c.CreateVolume(context.Background(), volReq)
+		Expect(err).NotTo(HaveOccurred())
+
+		By("creating snapshots")
+		snapshotReq := MakeCreateSnapshotReq(sc, "listSnapshots-snapshot-unrelated1", volumeUnrelated1.GetVolume().GetVolumeId())
```
xing-yang, line 1358: Comments

pkg/sanity/controller.go
```diff
@@ -1341,34 +1341,63 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *TestContext
...
+		snapshotUnrelated1, err := c.CreateSnapshot(context.Background(), snapshotReq)
+		Expect(err).NotTo(HaveOccurred())
+
+		snapshotReq = MakeCreateSnapshotReq(sc, "listSnapshots-snapshot-target", volumeTarget.GetVolume().GetVolumeId())
+		snapshotTarget, err := c.CreateSnapshot(context.Background(), snapshotReq)
 		Expect(err).NotTo(HaveOccurred())
 
+		snapshotReq = MakeCreateSnapshotReq(sc, "listSnapshots-snapshot-unrelated2", volumeUnrelated2.GetVolume().GetVolumeId())
```
xing-yang, line 1366: Comments

pkg/sanity/controller.go
xing-yang: Can you move this up to right after create volume.

timoreimann:
@xing-yang I addressed the comments.

However, the _"should return snapshots that match the specified source volume id"_ test still fails, apparently because the (mock) driver used in the tests does not support multiple snapshots sourced by the same volume ID.

xing-yang:
> However, the "should return snapshots that match the specified source volume id" test still fails, apparently because the (mock) driver used in the tests does not support multiple snapshots sourced by the same volume ID. I'm trying to understand if the driver needs to be extended or the test reduced.

Then the mock driver should be modified to handle that test.

timoreimann:
I simplified the test for now just to see if it passes before I modify the mock driver. Looking into that next.

pkg/sanity/controller.go
```diff
@@ -1340,35 +1340,71 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *TestContext
 	})
 
 	It("should return snapshots that match the specified snapshot id", func() {
+		// The test creates three snapshots: one that we intend to find by
+		// snapshot ID, and two unrelated ones that must not be returned by
```
xing-yang, line 1344: When you say "unrelated", do you mean they won't be returned when filtering by snapshot id or filtering by source volume id?  Please clarify in the comments.

timoreimann: As this is the test for snapshot IDs, the "unrelatedness" relates to filtering by snapshot IDs. I'll amend the comments.

timoreimann:
@xing-yang I addressed your last comment and brought back the version of the source volume ID test that expects multiple snapshots to be returned. The mock was updated accordingly.

Please take another look.

timoreimann:
Uh, I don't understand this: `make test` runs flawlessly for me locally but in the CI it fails. :(

timoreimann:
I don't think this is flakiness, but anyway:

/retest

timoreimann:
I think I understand now why the tests kept failing: I missed to see that the CI runs the tests both against the mocked as all as the hostpath driver, and the latter wasn't updated yet.

I briefly looked at what would need to be changed about the hostpath driver to return multiple snapshots. While the necessary changes are pretty clear, at this point I'd definitely favor getting the PR approved and merged in the simplified version that doesn't require changing any CSI drivers. I'd happy to create an issue to track the extension and work on it afterwards.

@xing-yang what do you think, does that sound like a reasonable approach to you?

xing-yang:
> I think I understand now why the tests kept failing: I missed to see that the CI runs the tests both against the mocked as all as the hostpath driver, and the latter wasn't updated yet.

> I briefly looked at what would need to be changed about the hostpath driver to return multiple snapshots. While the necessary changes are pretty clear, at this point I'd definitely favor getting the PR approved and merged in the simplified version that doesn't require changing any CSI drivers. (I updated the PR to reflect this accordingly.) I'd happy to create an issue to track the extension and work on it afterwards.

> @xing-yang what do you think, does that sound like a reasonable approach to you?

@timoreimann sounds good to me.

xing-yang:
The tests look good to me.
Can you clean up the comments in "Special notes for your reviewer" section?

timoreimann:
@xing-yang 👍 done

timoreimann:
This one still needs an /approve I suppose?

pkg/sanity/controller.go
```diff
@@ -1340,35 +1340,71 @@ var _ = DescribeSanity("ListSnapshots [Controller Server]", func(sc *TestContext
...
+		// ListSnapshots.
+
+		By("creating first unrelated snapshot")
+		// Create volume source and afterwards the first unrelated snapshot.
+		volReq := MakeCreateVolumeReq(sc, "listSnapshots-volume-unrelated1")
+		volumeUnrelated1, err := c.CreateVolume(context.Background(), volReq)
```
pohly, line 1350: As pointed out elsewhere (issue #260), all created resources like volumes must be registered with the `cl` instance, otherwise they can leak in case of an assertion failure.

Not sure whether you want to address that already in this PR or do it separately - up to you.

timoreimann: If you don't mind, I'd prefer to merge this one as-is and do the proper registration for resource cleanup in a follow-up PR. As noted in #260, I'd be happy to work on that.

PR #261, opened 2020-04-27
Centralize cleanup of created resources

### LINKED ISSUE #260: Failed tests leak CSI resources
The vast majority (all?) of the tests in csi-test do not seem to clean up resources properly when an assertion failure occurs. I'm not too familiar with Gomega, but my understanding is that once an `EXPECT` fails, the test execution comes to a halt immediately, so none of the cleanup logic usually done by the end of a test (e.g., deleting volumes or snapshots) gets a chance to execute.

This is a problem because the created CSI resources leaking this way may affect other tests to the extent that they may now fail as well even though they otherwise wouldn't. Specifically, a test that does not create any fixtures would now have left-over objects from the failed tests and thus fail too, aggravating identification of why a test run fails.

It'd be good to come up with a means to reliably clean up resources regardless of what the outcome of a test (and any of its assertions) is.

timoreimann:
**What type of PR is this?**
> /kind feature

This change revamps the way resources (like volumes and now also snapshots) are managed in tests with regards to cleaning up. Instead of putting the onus of cleaning up on the test author, we extend `Cleanup` to automatically (un-)register resources as they are being used.

`Cleanup` now exposes a single API that implements both `ControllerClient` and `NodeClient` to make it easier for all garbage collection-worthy requests to be funnelled through the new API. The way this is implemented in `Cleanup` is by embedding both `ControllerClient` and `NodeClient`, and proxying to the actual methods before registering cleanup tasks and returning the results.

Consequently, we can throw away large chunks of cleanup test code and unify all `{Controller,Node}Client` access to the `Cleanup` variable. In essence, this makes it much easier to do the right thing as a test author since each existing `Describe` context will provide a single
interaction point to the CSI APIs only.

For frequently used resource creation operations, we also provide `Must*` equivalents that fail the test if the results are unexpected. This makes our test code even more streamlined by DRYing out the number of assertions called.

List of other changes:

`cleanup.go`:
- Key volume and snapshot objects by ID instead of name. We have a few tests that omit or reuse the name, which makes it impossible to do automatic cleanup. Not printing the name of the resource as we clean up is a small price we have to pay for this adjustment, though.
- Fail tests when any cleanup operation errors out, except when we see error codes indicating that the resource is already cleaned up. Using a small logger wrapper to simplify automatic test failure.
- Rename `DeleteVolumes` to `Cleanup`.
- Provide convenience method `MustCreateSnapshotFromVolumeRequest` to create a sourcing volume and a snapshot in one go.

`controller.go`, `node.go`:
- Change all tests to use the API exposed by `Cleanup` only. (That is, do not offer `ControllerClient` and `NodeClient` directly anymore.)
- Register `Cleanup.Cleanup` in `AfterEach` where missing.
- Drop cleanup steps from various tests as this is now being taken care of by `Cleanup`.
- Use `Must*` equivalents were applicable.
- Use `HaveLen` to simplify length assertions.
- Make order of `Cleanup` variable initialization consistent.
- Minor cosmetic improvements.

Rename `Cleanup` to `Resources` and the file name accordingly.

Fixes #260

~`Cleanup` probably deserves a more generic name as this point, like `Resources`. I hesitated from renaming the variable (and the hosting file name) though to ease diffing the change. If this change and the rename proposal finds consensus, I'm happy to carry it out either through another commit or a follow-up PR.~ (As agreed on during the review, this PR now also does the rename.)

```release-note
Rename Cleanup to Resources and unexport cleanup (un-)registration, which is now handled implicitly and automatically.
```

/cc @pohly


pkg/sanity/cleanup.go
```diff
@@ -36,110 +42,284 @@ type VolumeInfo struct {
...
+// APIs. That way, Cleanup can ensure that resources are marked for cleanup as
+// necessary.
+// All methods can be called concurrently.
 type Cleanup struct {
-	Context                    *TestContext
-	ControllerClient           csi.ControllerClient
-	NodeClient                 csi.NodeClient
+	Context *TestContext
+	// ControllerClient is meant for struct-internal use only
+	csi.ControllerClient
```
pohly, line 54: It's meant for that, but we don't enforce that because the embedded member is getting exported, right?

I'm on the edge whether the API should prevent access by making it an unexpected member (`controllerClient csi.ControllerClient`). There may be valid cases where a user may want to call the methods that aren't wrapped.

I think I prefer keeping it like this.

pkg/sanity/cleanup.go
```diff
@@ -36,110 +42,284 @@ type VolumeInfo struct {
...
-					VolumeId:          info.VolumeID,
-					StagingTargetPath: cl.Context.StagingPath,
+				&csi.NodeUnpublishVolumeRequest{
+					VolumeId:   volumeID,
+					TargetPath: cl.Context.TargetPath + "/target",
 				},
 			); err != nil {
-				logger.Printf("warning: NodeUnstageVolume: %s", err)
+				if status.Code(err) != codes.NotFound {
+					Fail(fmt.Sprintf("NodeUnpublishVolume failed: %s", err))
```
pohly, line 260: We can't `Fail` during cleanup while there is still work to do. Instead the code should try to execute all operations, log and/or collect failures, and then in the end fail the test.

Otherwise cleaning up stops early although some other volumes perhaps could be deleted successfully.

pkg/sanity/cleanup.go
```diff
@@ -36,110 +42,284 @@ type VolumeInfo struct {
...
 }
 
-// UnregisterVolume removes the entry for the volume with the
-// given name, thus preventing all cleanup operations for it.
-func (cl *Cleanup) UnregisterVolume(name string) {
+// MustCreateSnapshot is like CreateSnapshot but asserts that the snapshot was
+// successfully created.
+func (cl *Cleanup) MustCreateSnapshot(ctx context.Context, req *csi.CreateSnapshotRequest) *csi.CreateSnapshotResponse {
+	snap, err := cl.createSnapshot(ctx, req)
+	Expect(err).NotTo(HaveOccurred())
```
pohly, line 177: A problem with assertions in helper functions is that Ginkgo only reports one source line for the error by default; even with `-trace`, that initial line is typically not very informative.

Better use `ExpectWithOffset` and add an offset parameter to the `MustCreateSnapshot` parameters so that this helper function an also be called indirectly through some other helper functions.

Also, `NotTo(HaveOccurred())` without additional explanation is potentially problematic, depending on how much information is in the error. Much too often the error is very generic, in which case the assertion produced by Gomega doesn't say anything about what failed.

Better always use `NoTo(HaveOccurred(), "create snapshot"`, potentially even with further parameters.

I know, much of the existing code doesn't do that properly either :cry:

pkg/sanity/cleanup.go
timoreimann: I left a comment only because unexporting wouldn't help: the (repo-internal) consumers are all part of the same `sanity` package, so they'd still be able to access `controllerClient`. It'd have to be moved into a sub-package, but I didn't want to go that far in my PR.

I can't think of too many reasons why users wouldn't want to go through the `Cleanup` layer: unless csi-test was buggy, it should do the right thing. The one case I can think of is when the `Cleanup()` part shouldn't be executed. Maybe that's reasonable, so let's stick to keeping it exported. I updated the comment to make the implications more clear.

pkg/sanity/cleanup.go
timoreimann: That's a good point. I created a small logger wrapper that tries to simplify the task.

pkg/sanity/cleanup.go
timoreimann: Ah yes, I remember you previously mentioned how `To()` should have a description but forgot about it again. I made sure it's now set.

I also updated the code to specify and/or pass through the offset everywhere. It's not a beauty though, I wonder if ginkgo could do better here by deriving the offset automatically (at least after the top-level `t.Helper()`-like indicator).

timoreimann:
@pohly all comments addressed, except for the two regarding the breaking change because I have one more dependent question:

I might not be fully familiar with how consumers of `csi-test` are expected to be given access. For the DigitalOcean CSI driver, we merely [configure and start the tests](https://github.com/digitalocean/csi-digitalocean/blob/59e354368961c4688243fc083c94b963c276e5b4/driver/driver_test.go#L97-L101). Do you (or others) go beyond that, specifically by accessing exported variables from the `Cleanup` / `Resources` struct?

The reason I'm asking is that I'm wondering if we should take advantage of the breaking change and start hiding parts of the package under an `internal` package. What are your thoughts on that?

pohly:
> Do you (or others) go beyond that, specifically by accessing exported variables from the Cleanup / Resources struct?

Yes, in PMEM-CSI we do have custom tests that are built on top of the sanity infrastructure, in addition to running the pre-defined tests: https://github.com/intel/pmem-csi/blob/018313154dff214da21fe39e6902d87857bc26e8/test/e2e/storage/sanity.go#L191-L230

> The reason I'm asking is that I'm wondering if we should take advantage of the breaking change and start hiding parts of the package under an internal package. What are your thoughts on that?

Please don't :sweat_smile:

timoreimann:
@pohly alrighty, I added a release note and moved v3 to v4 while also updating the links. Let me know if I missed something.

timoreimann:
Would love to see this getting approved and merged soonish because it touches a fair amount of existing tests, so other merges happening meanwhile stand a fair chance of generating merge conflicts.

pkg/sanity/logger.go
```diff
@@ -0,0 +1,59 @@
...
+	if err == nil {
+		return
+	}
+	l.failed = true
+	l.l.Printf(format, v...)
+}
+
+// Assert fails the spec if any error was logged.
+func (l *logger) Assert(offset int) {
+	ExpectWithOffset(offset+1, l.failed).To(BeFalse())
```
pohly, line 58: Directly calling https://pkg.go.dev/github.com/onsi/ginkgo?tab=doc#Fail with a caller skip parameter and a suitable message is probably going to look better in the resulting test failure.

If you want to make the failure message more informative, count errors and include the count here.

pohly: `defer logger.Assert()` looks odd. Assert - assert what?

Perhaps `logger.CheckForErrors()` or (similar to `framework.ExpectNoError`) `logger.ExpectNoErrors()`?

timoreimann: Implemented both.

timoreimann:
I rebased once more. Also noticed I missed some of the v3->v4 parts, so updated that as well.

It seems like I had to run `go mod vendor` as well. Did that with Go v1.12 as that seems to be the minimum Go version according to `go.mod`.

pohly:
> It seems like I had to run go mod vendor as well. Did that with Go v1.12 as that seems to be the minimum Go version according to go.mod.

I think go.mod specifies what we are compatible with. However, in practice this isn't getting tested: we only test with the Go version specified in https://github.com/kubernetes-csi/csi-test/blob/e89bc15a6e5e05e6adb4ed4f527e1ee3505d52a3/release-tools/travis.yml#L9

As long as Go 1.12 and 1.13 produce the same output, that doesn't matter. It worked here, so this is FYIO. However, I have seen cases where it didn't work and the pre-merge check with Go 1.13 complained.

This is true for all Kubernetes-CSI repos. I wonder whether we should:
* extend our testing to cover building and testing with several Go releases or
* bump up the version in go.mod.

The latter has the problem that it prevents downstream users from using an older Go even when that would still technically work. This is only an issue for repos that may get imported by others as a dependency (csi-test, csi-lib-utils).

@msau42: any thoughts on this?

pohly:
@timoreimann please rebase.

timoreimann:
@pohly rebased. I also verified that 1.13 does not differ with regards to `go mod tidy`ing / `vendor`ing, even though I think we figured that already.

timoreimann:
Hmm two snapshot tests failed. I'll look into that.

timoreimann:
FWIW, I think the [official Go policy is to support the two latest "major" versions](https://golang.org/doc/devel/release.html#policy) (not in semver terms). If we were to support multiple Go versions (which to my knowledge is somewhat common practice with consumable Go packages), then the policy is something we could model after, perhaps supporting three versions just to make the transition easier.

timoreimann:
The tests

```
  should return empty when the specified snapshot id does not exist
  /home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:1151
STEP: reusing connection to CSI driver at dns:///172.17.0.4:31890
• Failure [0.882 seconds]
ListSnapshots [Controller Server]
/home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
  should return empty when the specified snapshot id does not exist [It]
  /home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:1151
```

and

```
  should return empty when the specified source volume id does not exist
  /home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:1194
STEP: reusing connection to CSI driver at dns:///172.17.0.4:31890
• Failure [0.917 seconds]
ListSnapshots [Controller Server]
/home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/tests.go:44
  should return empty when the specified source volume id does not exist [It]
  /home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:1194
```

do **not** fail for me locally.

timoreimann:
The rerun succeeded. I wonder if this is some race condition with the new resource/cleanup logic, a problem with the (mock) driver, or something different. 🤔

pohly:
> any ideas? is this possibly a known issue?

Sorry, literally no idea. I am not familiar with those tests.

/test pull-kubernetes-csi-csi-test

pohly:
The test is now (?) flaky. I'm afraid until we understand why that happens we shouldn't merge this PR.

timoreimann:
Very understandable. I'll spend some time trying to figure out what the problem is.

Expect some dirty troubleshooting commits to this PR as I haven't been able to reproduce the problem locally.

timoreimann:
@pohly I figured out why the tests were failing: two `AfterEach()` blocks to clean up after listing and deleting snapshot tests were missing, so the left around snapshots affected other tests. I suppose it worked locally for me because of different execution orders.

I pushed a fixing commit and rebased from master. From my point of view, the PR is now good to move on.

PR #262, opened 2020-04-28
sanity: allow `Volume` `capacity_bytes` to be zero

### LINKED ISSUE #94: Test fails if CapacityBytes is always 0
According to the [CreateVolume specs](https://github.com/container-storage-interface/spec/blob/master/spec.md#createvolume) a volume should not set `CapacityBytes` (or set it to 0) to indicate that the capacity is unknown.
```
  // The capacity of the volume in bytes. This field is OPTIONAL. If not
  // set (value of 0), it indicates that the capacity of the volume is
  // unknown (e.g., NFS share).
  int64 capacity_bytes = 1;
```
When running the test `CreateVolume [Controller Server] [It] should fail when requesting to create a volume with already exisiting name and different capacity.` it fails as it is not possible to detect this if the volume capacity is always 0.

The test `CreateVolume [Controller Server] [It] should not fail when requesting to create a volume with already exisiting name and same capacity.` also fails but as far as I can see this can be fixed by setting `TestVolumeSize: 0`.

NicolasT:
**What type of PR is this?**
/kind bug

From the CSI spec:

> The capacity of the volume in bytes. This field is OPTIONAL. If not
> set (value of 0), it indicates that the capacity of the volume is
> unknown (e.g., NFS share).
> The value of this field MUST NOT be negative.

As such, the tests that validate the result of a `CreateVolume`
operation shall accept zero as the `capacity_bytes` of the returned
`Volume` object, when validating this value against the requested
capacity.

See: https://github.com/container-storage-interface/spec/blob/e129a75169c13bade7c7336afe21f259740ba679/spec.md#createvolume

Related to #94 but not fixing it completely (cfr. #263).

```release-note
The tests that validate the result of a `CreateVolume` operation now accept zero as the `capacity_bytes` of the returned `Volume` object, when validating this value against the requested
capacity.
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -482,7 +482,7 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 			Expect(vol.GetVolume()).NotTo(BeNil())
 			Expect(vol.GetVolume().GetVolumeId()).NotTo(BeEmpty())
 			cl.RegisterVolume(name, VolumeInfo{VolumeID: vol.GetVolume().GetVolumeId()})
-			Expect(vol.GetVolume().GetCapacityBytes()).To(BeNumerically(">=", TestVolumeSize(sc)))
+			Expect(vol.GetVolume().GetCapacityBytes()).To(Or(BeNumerically(">=", TestVolumeSize(sc)), BeZero()))
 
 			By("cleaning up deleting the volume")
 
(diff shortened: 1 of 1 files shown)


msau42:
/lgtm
/approve

Thanks for the fix! One last thing, can you add a release note for this fix? You can "/hold cancel" afterwards.
/hold

NicolasT:
Added a relnotes entry. Also, I removed the `Fixes` marker for #94 since this PR doesn't fix the whole issue (#263 plays a role as well).

/hold cancel

PR #265, opened 2020-05-06
Enforce offline volume expansion

gnufied:
Also Volume struct does not need mutex in it. No code is using it afaict.

/sig storage

### DIFF
--- a/mock/service/controller.go
+++ b/mock/service/controller.go
@@ -227,6 +227,11 @@ func (s *service) ControllerPublishVolume(
 	v.VolumeContext[ReadOnlyKey] = roVal
 	s.vols[i] = v
 
+	if volInfo, ok := MockVolumes[req.VolumeId]; ok {
+		volInfo.ISPublished = true
+		MockVolumes[req.VolumeId] = volInfo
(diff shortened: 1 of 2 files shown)


mock/service/controller.go
```diff
@@ -227,6 +227,11 @@ func (s *service) ControllerPublishVolume(
 	v.VolumeContext[ReadOnlyKey] = roVal
 	s.vols[i] = v
 
+	if volInfo, ok := MockVolumes[req.VolumeId]; ok {
+		volInfo.ISPublished = true
```
jsafrane, line 231: `ISPublished` is used for NodePublish/Unpublish, not for ControllerPublish. If you need to check for attached volumes before resizing them, add a separate IsControllerPublished flag to `volInfo`.

Which brings me to question, expanding an attached and unmounted volume is online or offline resize?

gnufied: it is still considered online.

gnufied: ack.  the offline check does uses `IsPublished` check, so I will update that code too.

PR #266, opened 2020-05-06
sanity: Add more input validation tests to node service

NicolasT:
**What type of PR is this?**
/kind feature

- Test for `NOT_FOUND` on `NodeStageVolume` and `NodeUnstageVolume` for unknown volumes
- Test for `NOT_FOUND` on `NodePublishVolume` for unknown volumes
- Rework the `NodePublishVolume` tests to not depend on input validation ordering inside the driver
- Test for `FAILED_PRECONDITION` on `NodePublishVolume` if `STAGE_UNSTAGE_VOLUME` is supported

```release-note
- The `csi-sanity` test suite now validates a nodeplugin returns `NOT_FOUND` when `NodeStageVolume` or `NodeUnstageVolume` is called given a non-existing `volume_id`.
- The `csi-sanity` test suite now validates a nodeplugin returns `NOT_FOUND` when `NodePublishVolume` is called given a non-existing `volume_id`.
- The `csi-sanity` test suite now validates a nodeplugin with the `STAGE_UNSTAGE_VOLUME` capability returns `FAILED_PRECONDITION` when `NodePublishVolume` is called given an empty `staging_target_path`.
```

### DIFF
--- a/hack/_apitest2/api_test.go
+++ b/hack/_apitest2/api_test.go
@@ -15,10 +15,10 @@ func TestMyDriverWithCustomTargetPaths(t *testing.T) {
 	var createTargetDirCalls, createStagingDirCalls,
 		removeTargetDirCalls, removeStagingDirCalls int
 
-	wantCreateTargetCalls := 3
-	wantCreateStagingCalls := 3
-	wantRemoveTargetCalls := 3
-	wantRemoveStagingCalls := 3
+	wantCreateTargetCalls := 4
+	wantCreateStagingCalls := 4
+	wantRemoveTargetCalls := 4
+	wantRemoveStagingCalls := 4
 
 	// tmpPath could be a CO specific directory under which all the target dirs
 	// are created. For k8s, it could be /var/lib/kubelet/pods under which the
--- a/hack/_apitest2/api_test.go
+++ b/hack/_apitest2/api_test.go
@@ -15,10 +15,10 @@ func TestMyDriverWithCustomTargetPaths(t *testing.T) {
 	var createTargetDirCalls, createStagingDirCalls,
 		removeTargetDirCalls, removeStagingDirCalls int
 
-	wantCreateTargetCalls := 4
-	wantCreateStagingCalls := 4
-	wantRemoveTargetCalls := 4
-	wantRemoveStagingCalls := 4
+	wantCreateTargetCalls := 5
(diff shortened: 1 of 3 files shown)


NicolasT:
Same applies to `NodePublishVolume`, apparently (https://github.com/container-storage-interface/spec/blob/87a27a2496565857ed564284eaa762587feb9453/spec.md#nodepublishvolume-errors). I'll add one more commit to this PR.

NicolasT:
Prow fails because the E2E tests run against the `csi-driver-hostpath` driver, which seems to be incompatible with the changes introduced in this PR (may one then conclude `csi-driver-hostpath` is not fully spec-compliant itself?)

May look into this. However, [as discussed on Slack](https://kubernetes.slack.com/archives/C09QZFCE5/p1588768374317600), I'm starting to believe the requirements in the spec for a node plugin to validate a given `volume_id` and assert the volume's existence may be a mistake in the first place...

pohly:
Nicolas Trangez <notifications@github.com> writes:
> Prow fails because the E2E tests run against the `csi-driver-hostpath`
> driver, which seems to be incompatible with the changes introduced in
> this PR (may one then conclude `csi-driver-hostpath` is not fully
> spec-compliant itself?)

It wouldn't be the first time that adding more tests reveals
deficiencies in the example driver.

NicolasT:
> It wouldn't be the first time that adding more tests reveals deficiencies in the example driver.

That happens :smiley: Is there any 'standard' process to get things in sync? It appears the tests in the hostpath driver repository rely on a released version of `csi-sanity`, so fixing the driver would be 'in the dark' until a new version of `csi-sanity` including the new test is made, which relies on the hostpath driver being compatible, etc...

pohly:
> Is there any 'standard' process to get things in sync? It appears the tests in the hostpath driver repository rely on a released version of csi-sanity, so fixing the driver would be 'in the dark' until a new version of csi-sanity including the new test is made, which relies on the hostpath driver being compatible, etc...

A developer will have to test locally that the updated driver passes testing with the updated csi-test, then we gradually have to roll out the changes: first merge the fixed into the driver, do a new release with it, then update csi-release-tools to test against that new release. At that time, the new test can be added to csi-test without test failures.

It's a lengthy process, but there's no way around that when different repos are involved.

PR #267, opened 2020-05-08
Update release-tools to support building on Mac OS X

timoreimann:
**What type of PR is this?**
> /kind feature

Building / running on tests on Mac OS X does not work. This change updates the release-tools to the kubernetes-csi/csi-release-tools@9084fec which already has support for native building/compiling.

```release-note
Update release-tools to 9084fec to support native building
```

### DIFF
--- a/release-tools/SIDECAR_RELEASE_PROCESS.md
+++ b/release-tools/SIDECAR_RELEASE_PROCESS.md
@@ -50,18 +50,22 @@ naming convention `<hostpath-deployment-version>-on-<kubernetes-version>`.
 ## Release Process
 1. Identify all issues and ongoing PRs that should go into the release, and
   drive them to resolution.
-1. Download [K8s release notes
+1. Download v2.8+ [K8s release notes
   generator](https://github.com/kubernetes/release/tree/master/cmd/release-notes)
 1. Generate release notes for the release. Replace arguments with the relevant
   information.
-    ```
-    GITHUB_TOKEN=<token> ./release-notes --start-sha=0ed6978fd199e3ca10326b82b4b8b8e916211c9b --end-sha=3cb3d2f18ed8cb40371c6d8886edcabd1f27e7b9 \
-    --github-org=kubernetes-csi --github-repo=external-attacher -branch=master -output out.md
-    ```
-    * `--start-sha` should point to the last release from the same branch. For
-    example:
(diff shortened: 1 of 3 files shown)


timoreimann:
/assign @pohly 

since we talked about this one a couple of months ago [on Slack](https://kubernetes.slack.com/archives/C8EJ01Z46/p1576494088015200?thread_ts=1576485706.010900&cid=C8EJ01Z46).

timoreimann:
All tests seem to pass, and yet the build is red. I can see

```
make: Target 'test' not remade because of errors.
ERROR: 'make test' failed
WARNING: 'make test' failed, proceeding anyway
```

being logged.

I wonder if the fact that I modified the release tooling is the problem? [This comment](https://github.com/timoreimann/csi-test/blob/0a373995519a755226528ba580c808669e91cc18/release-tools/prow.sh#L1031-L1033) hints at that. If that's true, how'd I go about making the necessary change?

pohly:
The `make test-subtree` failure is legitimate. This change must be merged into https://github.com/kubernetes-csi/csi-release-tools/ first, then the updated csi-release-tools needs to be merged into this (and other repos, as needed) via `git subtree` (https://github.com/kubernetes-csi/csi-release-tools/#sharing-and-updating).

timoreimann:
@pohly turned out kubernetes-csi/csi-release-tools already had support for building natively, so all it takes for us is to bump csi-test's subtree reference to the latest master. I updated the PR accordingly, please take a look again whenever you have a moment.

timoreimann:
Just to be clear because Github makes it look like I pushed more changes than I actually did: for this PR I invoked

```
git subtree pull --prefix=release-tools https://github.com/kubernetes-csi/csi-release-tools.git master
```

and pushed a single commit. I think Github's support for displaying subtree-based changes is still behind its capabilities to do the same in submodules cases.

pohly:
> turned out kubernetes-csi/csi-release-tools already had support for building natively

Would be nice if I had remembered that I already changed that, right? :stuck_out_tongue_closed_eyes:

ETOOMUCHGOINGON

> I think Github's support for displaying subtree-based changes is still behind its capabilities to do the same in submodules cases.

This has confused others in the past, too. The advantage of subtree is that it is easy to do a test PR in a component using csi-release-tools and then submit that same commit against csi-release-tools itself. Might be worthwhile to reconsider the tool choice, though.

PR #269, opened 2020-05-31
Raw volumes coverage improvements

taaraora:
Raw volumes were added to the next cases:
- ExpandVolume [Controller Server]
- CreateSnapshot [Controller Server]
- DeleteSnapshot [Controller Server]
- ListSnapshots [Controller Server]
- ControllerPublishVolume
- CreateVolume
    - from snapshot
    - from source volume

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

Fixes #

```release-note
--csi.testvolumeaccesstype=block is now also used by the following tests: ExpandVolume [Controller Server], CreateSnapshot [Controller Server], DeleteSnapshot [Controller Server], ListSnapshots [Controller Server], ControllerPublishVolume, Create Volume from snapshot, Clone Volume
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -416,7 +416,8 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 			Expect(serverError.Code()).To(Equal(codes.InvalidArgument))
 		})
 
-		It("should return appropriate values SingleNodeWriter NoCapacity Type:Mount", func() {
+		// TODO: whether CreateVolume request with no capacity should fail or not depends on driver implementation
+		It("should return appropriate values SingleNodeWriter NoCapacity", func() {
 
 			By("creating a volume")
 			name := UniqueString("sanity-controller-create-single-no-capacity")
@@ -454,7 +455,7 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 			cl.UnregisterVolume(name)
 		})
 
-		It("should return appropriate values SingleNodeWriter WithCapacity 1Gi Type:Mount", func() {
+		It("should return appropriate values SingleNodeWriter WithCapacity 1Gi", func() {
 
 			By("creating a volume")
 			name := UniqueString("sanity-controller-create-single-with-capacity")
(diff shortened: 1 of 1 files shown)


pkg/sanity/sanity.go
```diff
@@ -190,7 +190,7 @@ func NewTestConfig() TestConfig {
 		StagingPath:          os.TempDir() + "/csi-staging",
 		CreatePathCmdTimeout: 10 * time.Second,
 		RemovePathCmdTimeout: 10 * time.Second,
-		TestVolumeSize:       10 * 1024 * 1024 * 1024, // 10 GiB
+		TestVolumeSize:       1 * 1024 * 1024 * 1024, // 10 GiB
```
pohly, line 193: Comment also needs to be updated.

pohly:
/ok-to-test

Please add some short release note text to the PR description.

pkg/sanity/sanity.go
taaraora: I had not enough space on my k8s cluster and didn't find flag to set test volume size at that time, that's why I changed it. But this change is not intended to be in final patch and I finally found `--csi.testvolumesize` flag.
It is rolled back now.

taaraora:
@pohly release note text was added.

pkg/sanity/controller.go
```diff
@@ -416,7 +416,8 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 			Expect(serverError.Code()).To(Equal(codes.InvalidArgument))
 		})
 
-		It("should return appropriate values SingleNodeWriter NoCapacity Type:Mount", func() {
+		// TODO: whether CreateVolume request with no capacity should fail or not depends on driver implementation
```
pohly, line 419: The commit message mentions added new tests, but all I find in the diff are changes to existing tests?

In this case here, the test doesn't even get changed, just its name.

I'm confused, sorry.

taaraora: The idea of this PR is to make possible running all the above-mentioned test cases for raw volumes.
Currently when sanity tests are executed with the flag `--csi.testvolumeaccesstype=block` above mentioned test cases are still executed for `mount` volumes.
My patch is based on the point, that all those cases use `MakeCreateVolumeReq` function for volume creation and I've just made it to produce raw volume when tests are executed for raw volumes.

taaraora: @pohly In commit message I wanted to say that raw volumes support was added to all those tests.
I've checked that list of tests cases and sure that when tests are executed with `--csi.testvolumeaccesstype=block` flag, all calls which bear volume access type are constructed with proper volume access type.

taaraora:
I agree I've updated release note.

pohly:
@jsafrane do you want to approve?

PR #277, opened 2020-07-29
Add mock CSI driver deploy example

Jiawei0227:
This commit adds an example folder which contains a deploy instruction
of the full CSI mock driver including csi-provisioner, csi-resizer,
csi-snapshotter and csi-node-driver-registrar.

This helps to playaround with the CSI mock driver to understand the
functionality.

Tested on GCP.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake
/kind example

Fixes #

This is a demo deployment of mock csi driver with all the containers.

```release-note
Deployment manifests and examples are added for mock CSI driver under `mock/example`
```

### DIFF
--- a/mock/example/deploy/csi-mock-driver-deployment.yaml
+++ b/mock/example/deploy/csi-mock-driver-deployment.yaml
@@ -1,0 +1,126 @@
+# This YAML file contains the deployment of the CSIDriver
+# including node-driver-registrar, external-provisioner, external-snapshotter
+# external-resizer and mock-driver
+#
+# This YAML file also includes deployment of the CSIDriver, StorageClass
+# VolumeSnapshotClass for the mock driver.
+
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: csi-mockplugin
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: csi-mockplugin
+  template:
+    metadata:
+      labels:
+        app: csi-mockplugin
+    spec:
+      serviceAccount: csi-mock
+      containers:
+        - name: csi-provisioner
+          image: k8s.gcr.io/sig-storage/csi-provisioner:v1.6.0
+          args:
+            - "--csi-address=/csi/csi.sock"
+          imagePullPolicy: "IfNotPresent"
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+          # Normally privileged: true is only required for the driver container to perform 
+          # the mount operation. However, privileged container is necessary for systems with 
+          # SELinux, where non-privileged sidecar containers cannot access unix domain socket
+          # created by privileged CSI driver container. The other containers in this file being 
+          # granted privileged permission are for the same reason.
+          securityContext:
+            privileged: true
+        - name: node-driver-registrar
+          image: k8s.gcr.io/sig-storage/csi-node-driver-registrar:v1.3.0
+          args:
+            - --v=5
+            - --csi-address=/csi/csi.sock
+            - --kubelet-registration-path=/var/lib/kubelet/plugins/io.kubernetes.storage.mock/csi.sock
+          env:
+            - name: KUBE_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  apiVersion: v1
+                  fieldPath: spec.nodeName
+          securityContext:
+            privileged: true
+          volumeMounts:
+          - mountPath: /csi
+            name: socket-dir
+          - mountPath: /registration
+            name: registration-dir
+        - name: csi-resizer
+          image: k8s.gcr.io/sig-storage/csi-resizer:v0.5.0
+          args:
+            - "--v=5"
+            - "--csi-address=/csi/csi.sock"
+          securityContext:
+            privileged: true
+          volumeMounts:
+          - mountPath: /csi
+            name: socket-dir
+        - name: csi-snapshotter
+          image:  k8s.gcr.io/sig-storage/csi-snapshotter:v2.1.0
+          args:
+            - "--v=5"
+            - "--csi-address=/csi/csi.sock"
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+          securityContext:
+            privileged: true
+        - name: mock-driver
+          image: k8s.gcr.io/sig-storage/mock-driver:v3.1.0
+          args:
+            - "--disable-attach=true"
+          env:
+            - name: CSI_ENDPOINT
+              value: /csi/csi.sock
+          # The driver container has to be privileged in order to do mount operation
+          securityContext:
+            privileged: true
+          volumeMounts:
+            - name: socket-dir
+              mountPath: /csi
+      volumes:
+        - name: socket-dir
+          hostPath:
+            path: /var/lib/kubelet/plugins/io.kubernetes.storage.mock
+            type: DirectoryOrCreate
+        - name: registration-dir
+          hostPath:
+            path: /var/lib/kubelet/plugins_registry
+            type: Directory
+
+---
+apiVersion: storage.k8s.io/v1
+kind: CSIDriver
+metadata:
+  name: io.kubernetes.storage.mock
+spec:
+  attachRequired: false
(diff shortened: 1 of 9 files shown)


Jiawei0227:
I signed it

example/README.md
```diff
@@ -0,0 +1,28 @@
...
+```
+$ kubectl apply -f deploy/csi-mock-driver-rbac.yaml
+$ kubectl apply -f deploy/csi-mock-driver-deployment.yaml
+```
+
+This will deploy a `CSIDriver` called `io.kubernetes.storage.mock`. There will be a deployment `csi-mockplugin` being 
+installed in the default namespace. Correspondingly, you can find a pod with prefix `csi-mockplugin`. In the meantime,
+ a `StorageClass` called `test-csi-mock` will be generated along with a `VolumeSnapshotClass` called `csi-mock-snapclass`.
+
+ Note that if you are using version prior to 1.17, [snapshot-controller](https://github.com/kubernetes-csi/external-snapshotter#usage) will require to be manual installed.
```
msau42, line 19: prior to 1.17, there was no snapshot controller. I think maybe a more accurate note is to say that the k8s distribution is supposed to install the snapshot-controller, but if it doesn't, then you'll have to manually install it separately.

example/deploy/csi-mock-driver-deployment.yaml
```diff
@@ -0,0 +1,125 @@
...
+      app: csi-mockplugin
+  template:
+    metadata:
+      labels:
+        app: csi-mockplugin
+    spec:
+      serviceAccount: csi-mock
+      containers:
+        - name: csi-provisioner
+          image: gcr.io/k8s-staging-csi/csi-provisioner:canary
```
msau42, line 25: We're switching to k8s.gcr.io as our image repo. Can you use the images available there? Here's this list of image tags currently available: 

https://github.com/kubernetes/k8s.io/blob/master/k8s.gcr.io/images/k8s-staging-sig-storage/images.yaml

example/deploy/csi-mock-driver-deployment.yaml
```diff
@@ -0,0 +1,125 @@
...
+          args:
+            - "--csi-address=$(ADDRESS)"
+            - "--leader-election"
```
msau42, line 28: We can probably disable leader election for all the sidecars since the mock driver is not resilient to restarts anyway.

example/deploy/csi-mock-driver-deployment.yaml
```diff
@@ -0,0 +1,125 @@
...
+          hostPath:
+            path: /var/lib/kubelet/plugins/io.kubernetes.storage.mock
+            type: DirectoryOrCreate
+        - name: registration-dir
+          hostPath:
+            path: /var/lib/kubelet/plugins_registry
+            type: Directory
+
+---
+apiVersion: storage.k8s.io/v1
```
msau42, line 101: v1 was made available in 1.18. So we should document that this example requires 1.18+

example/pvc-test.yaml
```diff
@@ -0,0 +1,11 @@
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: example-pvc
+spec:
+  accessModes:
+    - ReadWriteMany
```
msau42, line 7: mock driver is closer to readwriteonce. I'm not even sure it can handle data persisting across a pod restart, but it definitely doesn't support sharing data across multiple nodes.

example/README.md
```diff
@@ -0,0 +1,28 @@
+# CSI Mock Driver Example
```
msau42, line 1: This may be better under the mock/ directory. We can point to it from the top level readme.

In the mock/readme.md, it would be also good to describe the high level limitations of the mock driver, such as:
* requires all the components to run on the same node
* only supports single node
* does not persist data across restarts

example/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+This folder contains an example manifest of deploying CSIDriver including `csi-driver-node-registrar`, `csi-provisioner`,
+`csi-resizer` and `csi-snapshotter` onto a cluster. For testing purpose, `csi-attcher` is not included. Thus,
```
msau42, line 4: csi-attacher

example/pvc-test.yaml
```diff
@@ -0,0 +1,11 @@
...
+  storageClassName: test-csi-mock
```
msau42, line 8: missing a storageclass yaml?

Jiawei0227: The storageclass is included in deploy/csi-mock-driver-deployment.yaml

example/pvc-test.yaml
Jiawei0227: Done

example/deploy/csi-mock-driver-deployment.yaml
Jiawei0227: Done

example/deploy/csi-mock-driver-deployment.yaml
Jiawei0227: Done

example/README.md
Jiawei0227: Done

example/README.md
Jiawei0227: Done

example/README.md
Jiawei0227: Done

example/deploy/csi-mock-driver-deployment.yaml
Jiawei0227: Done

mock/example/deploy/csi-mock-driver-deployment.yaml
```diff
@@ -0,0 +1,124 @@
...
+        app: csi-mockplugin
+    spec:
+      serviceAccount: csi-mock
+      containers:
+        - name: csi-provisioner
+          image: gcr.io/k8s-staging-sig-storage/csi-provisioner:v1.6.0
+          args:
+            - "--csi-address=$(ADDRESS)"
+          env:
+            - name: ADDRESS
```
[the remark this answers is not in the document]
msau42: ditto throughout

mock/example/deploy/csi-mock-driver-deployment.yaml
```diff
@@ -0,0 +1,124 @@
...
+            privileged: true
+          volumeMounts:
+          - mountPath: /csi
+            name: socket-dir
+        - name: csi-snapshotter
+          image:  gcr.io/k8s-staging-sig-storage/csi-snapshotter:v2.1.0
+          args:
+            - "--v=5"
+            - "--csi-address=$(ADDRESS)"
+            - "--leader-election=false"
```
msau42, line 72: I think we don't actually need to explicitly set this since the default is false

mock/example/pod-test.yaml
```diff
@@ -0,0 +1,29 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: web-server
+spec:
+  # deploy this pod only on where csi-mock-driver exists
+  # This is because csi-mock-driver keeps all states in memory and the process that
+  # provisioned the PV needs to be the same process that's mounting it
+  affinity:
```
msau42, line 9: as a future improvement, if we add topology support to the mock driver, then we don't need this in the pod. The pv provisioned will have a nodeaffinity on it, and the pod will always get scheduled to the correct node.

msau42:
/approve
lgtm, can you squash your commits into one?

msau42:
regarding you question on privileged vs running as uid 0, my basic understanding is that privileged enables a lot more than just the uid: all the host devices get mounted in the container, and there's also a number of linux capabilities that get enabled: https://kubernetes.io/docs/concepts/policy/pod-security-policy/#privileged

Jiawei0227:
Okay, thanks! Commits squashed.

msau42:
Can you also add a release note saying that deployment manifests and examples are added for mock driver

PR #279, opened 2020-08-03
Use klog for mock CSI driver logging

### LINKED ISSUE #278: mock CSI driver should log the timestamp of events
It seems right now, all the events being logged in CSI driver does not have a timestamp. This can make triage k8s e2e test failure quite hard.

/cc @msau42

Jiawei0227:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

This PR replaces logging information from mock CSI driver from fmt.Printf to klog

Fixes #278

After this goes to release, we might need to change other places where uses it to add "--v=5" to the startup args.

```release-note
Replace logging to klog for mock CSI driver.
```

/cc @msau42

### DIFF
--- a/cmd/mock-driver/main.go
+++ b/cmd/mock-driver/main.go
@@ -28,7 +28,15 @@ import (
 	"github.com/kubernetes-csi/csi-test/v4/driver"
 	"github.com/kubernetes-csi/csi-test/v4/mock/service"
 	"gopkg.in/yaml.v2"
+	"k8s.io/klog"
 )
+
+func init() {
+	// klog verbosity guide for this package
+	// Use V(2) for one time config information
+	// Use V(3) for general debug information logging
+	klog.InitFlags(flag.CommandLine)
+}
 
 func main() {
 	var config service.Config
@@ -56,7 +64,7 @@ func main() {
 		if err == nil {
 			config.ExecHooks = execHooks
 		} else {
-			fmt.Printf("Failed to load hooks file %s: %v", hooksFile, err)
+			klog.Errorf("Failed to load hooks file %s: %v", hooksFile, err)
 		}
 	}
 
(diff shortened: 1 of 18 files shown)


driver/driver.go
```diff
@@ -236,7 +236,7 @@ func logGRPC(method string, request, reply interface{}, err error) {
 	}
 
 	msg, _ := json.Marshal(logMessage)
-	fmt.Printf("gRPCCall: %s\n", msg)
+	klog.V(3).Infof("gRPCCall: %s\n", msg)
```
pohly, line 239: There is code in Kubernetes which parses this output: https://github.com/kubernetes/kubernetes/blob/ec560b9737537be8c688776461bc700e8ddedb9d/test/e2e/storage/csi_mock_volume.go#L1354-L1360

I suspect using klog here will break that.

I'm leaning towards doing this (and only this) output with `fmt.Printf`.

Jiawei0227: Yeah, that makes sense. Just wondering if there is anyway we can add timestamps to these logs without breaking the e2e tests? The context is that it might be good for us to understand when it receives the gRPCCall for triage some issues.

pohly: The alternative is to make theE2E tests more flexible such they support such lines with and without the standard klog prefix, then update the mock driver.

Jiawei0227: It seems the E2E test is using a static version of mock CSI driver which is v3.1.0. So unless we update the driver version there, we should not break the test. 

Should the right order be: 
1. Update the mock CSI driver to use log.
2. Update the e2e test to use latest driver version tag and also apply changes to the test cases?

pohly: I prefer to update the E2E test in advance, because then at some point in the future, the mock driver image can be updated without having to remember that the change is still needed.

Jiawei0227: https://github.com/kubernetes/kubernetes/pull/93950
PR is out.

cmd/mock-driver/main.go
```diff
@@ -28,8 +28,17 @@ import (
...
 	"gopkg.in/yaml.v2"
+	"k8s.io/klog"
 )
 
+func init() {
+	// klog verbosity guide for this package
+	// Use V(2) for one time config information
+	// Use V(3) for general debug information logging
+	klog.InitFlags(flag.CommandLine)
+	flag.Set("logtostderr", "true")
```
pohly, line 39: Setting logtostderr should be redundant: https://github.com/kubernetes/klog/blob/4f12f780d4781eb828bc6520ae87d7516f30f0b2/klog.go#L38-L40

Can you remove this line and double-check?

PR #297, opened 2020-09-29
Prevent plugin from returning in-use error for source volumes

timoreimann:
**What type of PR is this?**
> /kind bug

This PR addresses a regression in #261 causing plugins to return an in-use error (`FAILED_PRECONDITION`) when a sourcing resource (i.e., a snapshot or a volume) is deleted before the sourced volume is.

See https://github.com/kubernetes-csi/csi-test/pull/289#issuecomment-688339389 for details.

```release-note
Clean up sourced volumes prior to deleting sourcing resources
```

@apurv15 could you give this a try with your driver?


pkg/sanity/controller.go
```diff
@@ -587,7 +587,14 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
...
 			}
-			_, err := r.CreateVolume(context.Background(), vol2Req)
+			volume2, err := r.CreateVolume(context.Background(), vol2Req)
+			Expect(err).NotTo(HaveOccurred())
+
+			// Clean up sourced volume first to prevent plugin from returning an
+			// in-use error.
+			By("cleaning up deleting the volume created from snapshot")
+			delVol2Req := MakeDeleteVolumeReq(sc, volume2.GetVolume().GetVolumeId())
+			_, err = r.DeleteVolume(context.Background(), delVol2Req)
```
pohly, line 597: This looks like the cleanup code didn't delete volumes in the right order, correct?

I think we should fix that instead of adding custom cleanup code in the tests themselves. Suppose we add another check after creating the second volume and before deleting it: if that check then fails, test cleanup will fail like it does now.

pohly: I haven't checked how volumes are tracked, but perhaps clean up with last-in-first-out order?

timoreimann: We [currently always delete snapshots first, then volumes](https://github.com/kubernetes-csi/csi-test/blob/0ddc0e0dd7df52b42db71e48c890fd3c254add10/pkg/sanity/resources.go#L251-L252). Deletion in LIFO order sounds like the right approach to me, I think it should work. (For this particular test case, it means that we'd delete volume 2 first, then the snapshot, then the volume 1.)

I'll work on making the necessary updates to this PR.

timoreimann: I made the discussed adjustments. Will comment some of the changes via Github comments to add more clarity.

Please have a look and let me know what you think.

pkg/sanity/resources.go
```diff
@@ -26,19 +27,62 @@ import (
...
 
+	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
-// VolumeInfo keeps track of the information needed to delete a volume.
-type VolumeInfo struct {
+// managedResourceInfos is a map of managed resource infos keyed by resource
+// IDs.
+type resourceInfos map[string]resourceInfo
```
timoreimann, line 36: Alternatively, this could also be typed as `[]resourceInfo` which would make the LIFO property more natural/obvious. On the downside, it would also bring extra effort to remove items from the slice as we clean up resources (volumes, snapshots) during `Cleanup`.

I decided to go with a map and introduced an `order` field to keep track of when resources where added to the list.

pkg/sanity/resources.go
```diff
@@ -26,19 +27,62 @@ import (
...
+
+// resourceInfo represents a resource (i.e., a volume or a snapshot).
+type resourceInfo struct {
+	id    string
+	order int
+	data  interface{}
+}
+
+// volumeInfo keeps track of the information needed to delete a volume.
+type volumeInfo struct {
```
timoreimann, line 77: I unexported the type since it's effectively an implementation detail. I guess strictly speaking, this could be considered a breaking change, though I doubt anyone uses it (or rather hope not so).

If we're concerned about this, I'm also happy to undo the change and safe it for when we have more things to change that justify a new major release.

pkg/sanity/resources.go
```diff
@@ -26,19 +27,62 @@ import (
...
 	// Node on which the volume was published, empty if none
 	// or publishing is not supported.
 	NodeID string
-
-	// Volume ID assigned by CreateVolume.
-	VolumeID string
```
timoreimann, line 39: I removed `VolumeID` because it's now part of the key in `resourceInfos`. Same considerations as with the unexporting of `[Vv]olumeInfo` apply, though.

pkg/sanity/resources.go
pohly: This uses the volume ID and snapshot ID as key, right? Is is guaranteed that those never conflict?

> it would also bring extra effort to remove items from the slice as we clean up resources (volumes, snapshots) during Cleanup

That extra effort is just finding the item and then creating a slice without it, right? That doesn't seem that hard.

Without having seen what it looks like, I suspect that `[]resourceInfo` will be the better solution.

apurv15:
@timoreimann : We are seeing that a lot more test cases are failing with this change.
e.g.
ListSnapshots [Controller Server] 
  should return snapshots that match the specified snapshot id
  /root/csi-sanity-new/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:1126
STEP: reusing connection to CSI driver at /var/lib/kubelet/plugins/org.veritas.infoscale/csi.sock
STEP: creating first unrelated snapshot
STEP: creating target snapshot
STEP: creating second unrelated snapshot
STEP: listing snapshots
cleanup volume: deleting vol_c0o3m11862x7xmlbt5o9
cleanup volume: DeleteVolume failed: rpc error: code = FailedPrecondition desc = Could not delete volume vol_c0o3m11862x7xmlbt5o9, as it has dependent snapshots ['snap_043b9dl9u20l60xj4f2i'] with sync in progress

pkg/sanity/resources.go
timoreimann: Made the change. I think you were right, this looks much simpler.

timoreimann:
@apurv15 I'm not exactly sure why things don't work for you. 🤔 

I pushed another commit that outputs extra debug messages when you set the `DEBUG_RESOURCE_MANAGER` environment variable to something non-empty. Could you try this out and share the log output, especially the `[DEBUG]`-prefixed messages?

@pohly let me know if we think it makes sense to keep the poor man's debug logging approach in or not.

pohly:
>  let me know if we think it makes sense to keep the poor man's debug logging approach in or not.

Why not just `klog.V(4)` or so?

timoreimann:
This one fell off my radar. I'll make sure to address the last comment soon.

timoreimann:
@pohly 

>Why not just `klog.V(4)` or so?

We currently [set a `log.Logger` to write into a `GinkgoWriter`](https://github.com/kubernetes-csi/csi-test/blob/054980205e4fc45ab9e940e39bdfd71066d96d87/pkg/sanity/logger.go#L33), causing the output to be suppressed if all tests pass. Unless I'm missing something, klog only supports [setting the output at the global package level](https://godoc.org/github.com/kubernetes/klog#SetOutput). I'm a bit reluctant to do that as it'd mean that we could overwrite klog settings that csi-test library consumers may have set for their tests.

WDYT?

pohly:
> We currently set a log.Logger to write into a GinkgoWriter, causing the output to be suppressed if all tests pass.

We could do the same *in the csi-sanity* binary and enable the klog flags, which will have the same effect. We can't do it when used as lib, but then users who care can do something similar in their main function.

That should go a long way towards nicer logging without doing anything special (like a `DEBUG_RESOURCE_MANAGER` that most users won't know about).

pkg/sanity/resources.go
```diff
@@ -222,107 +211,119 @@ func (cl *Resources) registerSnapshot(offset int, id string) {
...
-// snapshots.
+// Cleanup calls unpublish methods as needed and deletes all managed resources.
 func (cl *Resources) Cleanup() {
 	cl.mutex.Lock()
 	defer cl.mutex.Unlock()
 	ctx := context.Background()
 
-	cl.deleteSnapshots(ctx, 2)
-	cl.deleteVolumes(ctx, 2)
+	// Clean up resources in LIFO order to account for dependency order.
```
xing-yang, line 245: PVC and VolumeSnapshot have independent life cycles in Kubernetes.  If we fix the sanity test to make sure snapshots are always deleted before volumes are deleted, driver maintainers won't know it is possible for users to delete PVCs first and may run into the problems later on in production.

So I think we should actually have sanity tests that delete snapshot before and after the volume's deletion to make sure a driver can handle this independent life cycle.  Yes, this means driver will have to implement logic to do soft deletion if volume and snapshot are dependent to each other on the storage system in order to pass the sanity tests.

apurv15: I had commented on https://github.com/kubernetes-csi/csi-test/pull/289
Spec provides flexibility for PVC and its snapshots to be dependent or not. Spec https://github.com/container-storage-interface/spec/blob/master/spec.md says following:

===
CSI plugins SHOULD treat volumes independent from their snapshots.

If the Controller Plugin supports deleting a volume without affecting its existing snapshots, then these snapshots MUST still be fully operational and acceptable as sources for new volumes as well as appear on ListSnapshot calls once the volume has been deleted.

When a Controller Plugin does not support deleting a volume without affecting its existing snapshots, then the volume MUST NOT be altered in any way by the request and the operation must return the **FAILED_PRECONDITION** error code and MAY include meaningful human-readable information in the status.message field.
====

So, the test case should not validate just one behavior as the spec allows volumes and snapshots to be dependent and that is the case for some of the drivers.

timoreimann: With LIFO-ordered deletion implemented by this PR, I _think_ we should satisfy the minimum requirement that the spec mandates, which is that plugins do not or cannot treat volumes and snapshots independently. That is, if a volume is created from a snapshot, we first delete the snapshot and then the volume. (I also assume the same logic should apply if a volume is sourced from another volume.)

However, @apurv15 mentioned in [a previous comment](https://github.com/kubernetes-csi/csi-test/pull/297#issuecomment-705973373) that csi-test built from this change would cause more tests to fail. 🤔 I guess I should give the PR another hard to look to see if I missed anything. (@apurv15, did you ever get around running it with the enhanced debug log messages as I suggested?)

@xing-yang would it make sense to file a separate issue to track creating a new test for the missed scenario you mentioned (while taking into account the specifics of the spec as @apurv15 quoted above)? IMHO, this PR should focus on addressing the regression I introduced and returning to the [original cleanup order](https://github.com/kubernetes-csi/csi-test/blob/4b9faee268d03e35c94499387aeb4cdb532f4e58/pkg/sanity/controller.go#L699-L712).

timoreimann:
@pohly I replaced the custom logger and env var flag with klog as discussed. Also optimized / reduced the amount of logging a bit more.

Please take another look.

timoreimann:
I should manually test-drive this PR a bit more, so for now

/hold

pkg/sanity/resources.go
```diff
@@ -299,31 +296,33 @@ func (cl *Resources) cleanupVolume(ctx context.Context, offset int, volumeID str
 				Secrets:  cl.Context.Secrets.ControllerUnpublishVolumeSecret,
 			},
 		)
-		logger.Errorf(err, "ControllerUnpublishVolume failed: %s", err)
+		ExpectWithOffset(offset, err).NotTo(HaveOccurred(), "[volume cleanup] ControllerUnpublishVolume failed")
```
pohly: Previously errors during cleanup were logged, but cleanup continued, right? Now it aborts.

Error handling during cleanup is tricky. Ideally, as much as possible should get cleaned up and then a summary of all unsolved problems should be raised as a test failure. The latter ensures that a test that itself ran fine but then has problems during cleanup is reported as failed overall.

timoreimann: Whoops, somehow I missed that we used to continue to clean up even when we hit errors.

I made another update to collect all errors and assert at the end of the `Cleanup` call. I think this goes even beyond what we had before, where we'd assert at the end of a `cleanup{Volume,Snapshot}` call.

One thing I want to mention is that failure to, say, node-unpublish a volume, could mean that follow-up cleanup operations on the same resource (node-unstage, controller-unpublish, delete) may then be more likely to fail as well. I'm wondering if this aggravates understanding the produced error messages, and whether it could make more sense to not try follow-up cleanup operations within a single resource (but continue to try to clean up all resources).

I followed your suggestion for now, but let me know if you have thoughts on the above and whether it should influence another update.

pohly: I agree that errors are likely to cause other errors. But that is nothing new, developers should be familiar with follow-up errors and focused on the first one when trying to understand what went wrong.

So I think this behavior is fine.

pohly (approved):
/lgtm

pkg/sanity/resources.go
xing-yang: There are also storage systems that cannot delete snapshots before deleting PVCs. Our existing e2e tests also delete in a specific order, deleting snapshots first and then deleting PVCs. However, that caused some CSI drivers to fail because they require an opposite ordering of deletion which is PVCs must be deleted before snapshots are deleted.  There's an issue opened to track and fix this in e2e tests so that deletion of PVC and snapshots should be independent of each other:
https://github.com/kubernetes/kubernetes/issues/96719

timoreimann: @xing-yang how should csi-test (and the e2e tests) account for the two possible orderings?

Should we try one order, and if we receive a particular error, try the inverse?

Or is the expectation that the driver acknowledges delete requests and completes the cleanup asynchronously once both the snapshot and volume deletes have been issued?

csi-test currently does not check that deletes have completed, so we might naturally support the latter scenario but not the former.

xing-yang: We should test both cases:
- Delete volumes before deleting snapshots
- Delete snapshots before deleting volumes

CSI driver should pass both tests.  This means CSI driver will need to keep reference count internally and allow deletion to happen successfully regardless of the delete ordering.

xing-yang: CC @msau42

pohly: This doesn't need to be solved in this PR. It's more like an idea for some dedicated tests.

pohly: Regarding cleanup order: of both cases are valid, then picking one for cleanup (LIFO at the moment) is correct, i.e. we can merge this PR.

xing-yang: Regarding this comment `Clean up resources in LIFO order to account for dependency order.`, we should remove the text "to account to dependency order" as we are not trying to make sure snapshots are deleted before volumes are deleted.

timoreimann: In its current form, the code cleans up resources "backwards": the resource (either volume or snapshot) that was created last is cleaned up first, followed by cleaning up the resource that was created second-to-last, and so on. Given that sourcing volumes must exist before the snapshots they are created from, this means that we first delete a snapshot followed by its sourcing volume.

Doesn't that imply that we _do_ make sure that snapshots are deleted before volumes are @xing-yang?

Or are we talking about the scenario where a volume is being created from a snapshot and a possible dependency in that chain?

pohly: We discussed this a bit further on Monday. @msau42 pointed out that it is valid when a CSI driver refuses to delete PVCs that have snapshots. It is *also* valid to refuse deleting snapshots when there is still a PVC. In other words, a CSI driver only needs to support *one* order of deletions, but not both.

This means that picking LIFO order will work for some drivers, but not all.

The right approach is to retry deletion after it failed once, i.e. try to delete snapshot, delete pvc, then if snapshot deletion failed, try again. I think this can be added relatively easily by iterating twice over all resources instead of just once.

We also discussed writing tests that explicitly cover both deletion orders. One test can do "delete snapshot, delete pvc, delete snapshot again (if necessary)", the other "delete pvc, delete snapshot, delete pvc again (if necessary)". This then works with all drivers and ensures that both is tested. But this should be added in a separate PR.

xing-yang:
I'm fine to have this PR merged first and have another PR to handle the volume/snapshot deletion ordering issue.

timoreimann:
If no one plans to, I can drive the follow up PR.

Thanks everyone involved.

/hold cancel

PR #303, opened 2020-11-11
Create target directory if it doesn't exist

### LINKED ISSUE #304: Verify that publish targetPath exists after NodePublish
SPs are responsible for making the targetPath available during NodePublish. We should validate that the the targetPath exists after a success. 

Our mock driver doesn't create the publish path and was depending on the Kubernetes bug to do it.

/kind feature
/help

huffmanca:
**What type of PR is this?**
/kind bug

After https://github.com/kubernetes/kubernetes/pull/88759/files we need to create the staging and target directories if they don't exist.

Fixes #304

```release-note
Updates the mock driver to create the target directory in `NodePublishVolume`, and remove it in `NodeUnpublishVolume`.
```

### DIFF
--- a/mock/service/node.go
+++ b/mock/service/node.go
@@ -203,15 +203,19 @@ func (s *service) NodePublishVolume(
 			ISEphemeral: true,
 		}
 	} else {
-		if req.GetStagingTargetPath() != "" {
-			exists, err := checkTargetExists(req.GetStagingTargetPath())
+		if req.GetTargetPath() != "" {
+			exists, err := checkTargetExists(req.GetTargetPath())
 			if err != nil {
 				return nil, status.Error(codes.Internal, err.Error())
 			}
 			if !exists {
-				status.Errorf(codes.Internal, "staging target path %s does not exist", req.GetStagingTargetPath())
+				// If target path does not exist we need to create the directory where volume will be staged
+				if err = os.Mkdir(req.TargetPath, os.FileMode(0755)); err != nil {
+					msg := fmt.Sprintf("NodePublishVolume: could not create target dir %q: %v", req.TargetPath, err)
+					return nil, status.Error(codes.Internal, msg)
+				}
 			}
-			v.VolumeContext[nodeMntPathKey] = req.GetStagingTargetPath()
+			v.VolumeContext[nodeMntPathKey] = req.GetTargetPath()
 		} else {
 			v.VolumeContext[nodeMntPathKey] = device
 		}
@@ -258,6 +262,12 @@ func (s *service) NodeUnpublishVolume(
 		// Check to see if the volume has already been unpublished.
 		if v.VolumeContext[nodeMntPathKey] == "" {
 			return &csi.NodeUnpublishVolumeResponse{}, nil
+		}
+
+		// Delete any created paths
+		err := os.RemoveAll(v.VolumeContext[nodeMntPathKey])
+		if err != nil {
+			return nil, status.Errorf(codes.Internal, "Unable to delete previously created target directory")
 		}
 
 		// Unpublish the volume.


mock/service/node.go
```diff
@@ -51,7 +51,11 @@ func (s *service) NodeStageVolume(
 		return nil, status.Error(codes.Internal, err.Error())
 	}
 	if !exists {
-		status.Errorf(codes.Internal, "staging target path %s does not exist", req.StagingTargetPath)
+		// If target path does not exist we need to create the directory where volume will be staged
+		if err = os.MkdirAll(req.StagingTargetPath, os.FileMode(0755)); err != nil {
```
msau42, line 55: According to the csi spec, the CO is responsible for creating the staging path. So we should not mkdir for nodestage.

pkg/sanity/sanity.go
```diff
@@ -285,15 +285,10 @@ func (sc *TestContext) Setup() {
 
 	By("creating mount and staging directories")
 
-	// If callback function for creating target dir is specified, use it.
-	targetPath, err := createMountTargetLocation(sc.Config.TargetPath, sc.Config.CreateTargetPathCmd, sc.Config.CreateTargetDir, sc.Config.CreatePathCmdTimeout)
```
msau42, line 289: I think we still need these because the staging directory is the responsibility of the CO, and the publish path parent directory is the responsibility of the CO

mock/service/node.go
huffmanca: I've returned this back to the original code.

pkg/sanity/sanity.go
huffmanca: Since we're no longer creating the staging path in NodeStageVolume, I've returned creation of this path to the sanity test; however, if we attempt to initialize the targetPath it will fail since this is completed during the NodePublishVolume calls...

huffmanca: We discussed this offline - after adjusting `NodeUnpublishVolume` to remove any created directories the tests pass without modification. This section has been reverted.

mock/service/node.go
```diff
@@ -203,15 +204,20 @@ func (s *service) NodePublishVolume(
...
-			exists, err := checkTargetExists(req.GetStagingTargetPath())
+		if req.GetTargetPath() != "" {
+			exists, err := checkTargetExists(req.GetTargetPath())
 			if err != nil {
 				return nil, status.Error(codes.Internal, err.Error())
 			}
 			if !exists {
-				status.Errorf(codes.Internal, "staging target path %s does not exist", req.GetStagingTargetPath())
+				// If target path does not exist we need to create the directory where volume will be staged
+				if err = os.MkdirAll(req.TargetPath, os.FileMode(0755)); err != nil {
```
msau42, line 214: I think this should just be mkdir. CO is supposed to create the parent

mock/service/node.go
```diff
@@ -254,12 +260,22 @@ func (s *service) NodeUnpublishVolume(
...
 		nodeMntPathKey := path.Join(s.nodeID, req.TargetPath)
+		nodeMntPathCreated := nodeMntPathKey + "Created"
 
 		// Check to see if the volume has already been unpublished.
 		if v.VolumeContext[nodeMntPathKey] == "" {
 			return &csi.NodeUnpublishVolumeResponse{}, nil
 		}
 
+		// Check to see if we created the volume, and if so delete it
+		if v.VolumeContext[nodeMntPathCreated] == "true" {
```
msau42, line 271: do we need this new key or should we always delete?

mock/service/node.go
huffmanca: Updated.

mock/service/node.go
huffmanca: I was initially hesitant on always deleting, but considering that this driver is used for testing I don't think there's an issue. I've adjusted this section to always delete the targetPath.

pohly:
The part that isn't clear to me is which tests are run against the CSI mock driver and why. https://github.com/kubernetes/kubernetes/blob/dd4560370743e94c9d5a477370bf25ffb49e973e/test/e2e/storage/csi_volumes.go#L31 explicitly mentions that the testsuites tests are not run against the mock diver.

msau42:
we're specifically looking at test cases in https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/csi_mock_volume.go, which tests imitates different permutations of csi drivers.

mock/service/node.go
```diff
@@ -254,12 +258,20 @@ func (s *service) NodeUnpublishVolume(
 		// nodeMntPathKey is the key in the volume's attributes that is set to a
 		// mock mount path if the volume has been published by the node
 		nodeMntPathKey := path.Join(s.nodeID, req.TargetPath)
+		nodeMntPathCreated := nodeMntPathKey + "Created"
```
msau42, line 261: this can be removed?

mock/service/node.go
```diff
@@ -254,12 +258,20 @@ func (s *service) NodeUnpublishVolume(
...
 
 		// Check to see if the volume has already been unpublished.
 		if v.VolumeContext[nodeMntPathKey] == "" {
 			return &csi.NodeUnpublishVolumeResponse{}, nil
 		}
 
+		// Delete any created paths
+		delete(v.VolumeContext, nodeMntPathCreated)
+		err := os.RemoveAll(v.VolumeContext[nodeMntPathKey])
```
msau42, line 270: The else case is odd: https://github.com/kubernetes-csi/csi-test/pull/303/files#diff-4690af0a8d51b3daac5423ee65ecc1107da2f28aa20bed0dad99ec2e9fd1c6b5R219

I opened up https://github.com/kubernetes-csi/csi-test/issues/305 to follow up on this

mock/service/node.go
huffmanca: This has been removed.

msau42:
/lgtm
/approve

/hold
Can you update the PR title, and also add a release note?

huffmanca:
@msau42 

The title has been updated, a release note added, and the created issue linked.

PR #308, opened 2020-11-12
Add kubelet directories to mock driver manifest

msau42:
**What type of PR is this?**
>
> /kind api-change

/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

After #303, we need to mount the kubelet pod directory so we can create the target path. Also fix a bug with validating the csi staging directory.

Fixes #

```release-note
Add kubelet pod and csi plugin directories to mock driver manifest
```

### DIFF
--- a/mock/example/deploy/csi-mock-driver-deployment.yaml
+++ b/mock/example/deploy/csi-mock-driver-deployment.yaml
@@ -86,9 +86,11 @@ spec:
           securityContext:
             privileged: true
         - name: mock-driver
-          image: k8s.gcr.io/sig-storage/mock-driver:v3.1.0
+          image: k8s.gcr.io/sig-storage/mock-driver:v4.1.0
           args:
             - "--attach-limit=50"
+            # Required for e2e test log parsing
+            - "--v=5"
           env:
             - name: CSI_ENDPOINT
               value: /csi/csi.sock
@@ -98,6 +100,10 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
+            - name: kubelet-pods-dir
+              mountPath: /var/lib/kubelet/pods
+            - name: kubelet-csi-dir
+              mountPath: /var/lib/kubelet/plugins/kubernetes.io/csi
       volumes:
         - name: socket-dir
           hostPath:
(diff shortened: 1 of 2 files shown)


mock/example/deploy/csi-mock-driver-deployment.yaml
```diff
@@ -88,6 +88,8 @@ spec:
           volumeMounts:
             - name: socket-dir
               mountPath: /csi
+            - name: kubelet-pods-dir
+              mountPath: /var/lib/kubelet/pods
```
pohly, line 92: A short comment here why `mountPropagation: Bidirectional` isn't needed would have been useful ("csi-mock driver creates a directory instead of mounting something").

Why don't we also need `/var/lib/kubelet/plugins/kubernetes.io/csi`? The mock driver doesn't need to create anything there, but it does check for existence:
https://github.com/huffmanca/csi-test/blob/16b637af5ddd1e3908cf81bb8dd3b3b9c5a14b40/mock/service/node.go#L49

pohly: I'm not sure why, but it works despite that missing directory:
```
I0222 13:33:07.205639       1 driver.go:239] gRPCCall: {"Method":"/csi.v1.Node/NodePublishVolume","Request":{"volume_id":"4","staging_target_path":"/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-d3a93b29-d135-48c2-8b29-6f13004febbc/globalmount","target_path":"/var/lib/kubelet/pods/4a044aa9-963c-4628-9aed-10c02dcba18c/volumes/kubernetes.io~csi/pvc-d3a93b29-d135-48c2-8b29-6f13004febbc/mount","volume_capability":{"AccessType":{"Mount":{"fs_type":"ext4"}},"access_mode":{"mode":1}},"volume_context":{"name":"pvc-d3a93b29-d135-48c2-8b29-6f13004febbc","storage.kubernetes.io/csiProvisionerIdentity":"1614000761660-8081-io.kubernetes.storage.mock"}},"Response":{},"Error":"","FullError":null}
```

@msau42 do you want me to merge this PR or investigate?

pohly: Another observation: `-v=5` is needed in the `mock/example/deploy/csi-mock-driver-deployment.yaml` file to get gRPC calls logged. Making that output depend on the verbosity level was one of the changes that went into the driver since the older releases.

msau42: It worked because there was a bug in the driver that didn't actually return an error upon failure. I've fixed that in the latest update, as well as adding the necessary arguments and comments.

pohly (approved):
/lgtm

PR #318, opened 2021-01-28
Add external-attacher to mock CSI driver deployment

chrishenzie:
**What type of PR is this?**
/kind feature

This PR bundles the [external-attacher](https://github.com/kubernetes-csi/external-attacher) in the mock CSI driver deployment manifest. Including this pod makes it easier for users to quickly test volume attachment / detachment behavior.

```release-note
NONE
```

/sig storage
/assign @msau42 
/assign @Jiawei0227


pohly:
> Including this pod makes it easier for users to quickly test volume attachment / detachment behavior.

Do users really deploy the mock driver to test attach/detach? Who are those users and why do they use the mock driver for this?

chrishenzie:
I'm one of those users. I've been making changes to the CSIAttacher and AttachDetachController and am using this mock driver deployment combined with E2E tests to gather performance metrics.

msau42:
One of the main purposes of mock driver is for Kubernetes developers to test changes to Kubernetes/sidecars. So I think we should be supporting use of all Kubernetes features.

pohly:
I don't mind adding it, I was just wondering. When I test sidecars, I do it with the hostpath driver, therefore I was wondering why someone would prefer to do it with the mock drivver.

PR #338, opened 2021-04-28
Extend NodeUnpublish test to verify cleanup of TargetPath (#336)

### LINKED ISSUE #336: Extend NodeUnpublish test to verify cleanup of node_publish path
It is expected that CSI driver will be responsible for removal of node_publish path. We don't have any sanity tests that verify this behaviour.

dobsonj:
**What type of PR is this?**
/kind bug

https://github.com/kubernetes/kubernetes/pull/101441 deprecates removal of the CSI NodePublish target_path by the kubelet. This must be done by the CSI plugin according to the CSI spec.
This PR adds a sanity test to verify that the CSI plugin removes target_path as part of NodeUnpublish.

Fixes #336 

Yes
```release-note
Removal of the CSI nodepublish path by the kubelet is deprecated. This must be done by the CSI plugin according to the CSI spec.
```

### DIFF
--- a/cmd/csi-sanity/main.go
+++ b/cmd/csi-sanity/main.go
@@ -81,6 +81,8 @@ func main() {
 	stringVar(&config.RemoveTargetPathCmd, "removemountpathcmd", "Command to run for target path removal")
 	stringVar(&config.RemoveStagingPathCmd, "removestagingpathcmd", "Command to run for staging path removal")
 	durationVar(&config.RemovePathCmdTimeout, "removepathcmdtimeout", "Timeout for the commands to remove target and staging paths, in seconds")
+	stringVar(&config.CheckPathCmd, "checkpathcmd", "Command to run to check a given path. It must print 'file', 'directory', 'not_found', or 'other' on stdout.")
+	durationVar(&config.CheckPathCmdTimeout, "checkpathcmdtimeout", "Timeout for the command to check a given path, in seconds")
 	stringVar(&config.SecretsFile, "secrets", "CSI secrets file")
 	stringVar(&config.TestVolumeAccessType, "testvolumeaccesstype", "Volume capability access type, valid values are mount or block")
 	int64Var(&config.TestVolumeSize, "testvolumesize", "Base volume size used for provisioned volumes")
--- a/pkg/sanity/node.go
+++ b/pkg/sanity/node.go
@@ -449,6 +449,64 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
 			Expect(ok).To(BeTrue())
 			Expect(serverError.Code()).To(Equal(codes.InvalidArgument), "unexpected error: %s", serverError.Message())
 		})
+
+		It("should remove target path", func() {
+			// This test may break for consumers that are using
+			// custom target path functions if they have not yet
+			// implemented similar functionality to check if the
+			// path exists. Skip this test if there is a custom
+			// command or function provided to create the path,
+			// but not yet provided to check the path.
+			if sc.Config.CreateTargetPathCmd != "" && sc.Config.CheckPathCmd == "" {
+				Skip("CreateTargetPathCmd was set, but CheckPathCmd was not. Please update your testing configuration to enable CheckPathCmd.")
+			}
+			if sc.Config.CreateTargetDir != nil && sc.Config.CheckPath == nil {
+				Skip("CreateTargetDir was set, but CheckPath was not. Please update your testing configuration to enable CheckPath.")
+			}
+
+			name := UniqueString("sanity-node-unpublish-volume")
+			vol := createVolume(name)
+			volid := vol.GetVolume().GetVolumeId()
+			volpath := sc.TargetPath + "/target"
+
+			By("Getting a node id")
+			nid, err := r.NodeGetInfo(
+				context.Background(),
+				&csi.NodeGetInfoRequest{})
+			Expect(err).NotTo(HaveOccurred())
+			Expect(nid).NotTo(BeNil())
+			Expect(nid.GetNodeId()).NotTo(BeEmpty())
+
+			By("Staging and publishing a volume")
+			conpubvol := controllerPublishVolume(name, vol, nid)
+			_ = nodeStageVolume(name, vol, conpubvol)
+			_ = nodePublishVolume(name, vol, conpubvol)
+
+			// Verify that the path exists before calling
+			// NodeUnpublishVolume.
+			By("Checking the target path exists")
+			pa, err := CheckPath(volpath, sc.Config)
+			Expect(err).NotTo(HaveOccurred(), "checking path %q", volpath)
+			Expect(pa).NotTo(Equal(PathIsNotFound), "path %q should have been created by CSI driver and the test config should be enabling testing for that path", volpath)
+
+			By("Unpublishing the volume")
+			_, err = r.NodeUnpublishVolume(
+				context.Background(),
+				&csi.NodeUnpublishVolumeRequest{
+					VolumeId:   volid,
+					TargetPath: volpath,
+				},
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			// The CSI spec states that the SP MUST delete
+			// the file or directory it created at this path
+			// as part of NodeUnpublishVolume.
+			By("Checking the target path was removed")
+			pa, err = CheckPath(volpath, sc.Config)
+			Expect(err).NotTo(HaveOccurred(), "checking path %q", volpath)
+			Expect(pa).To(Equal(PathIsNotFound), "path %q should have been removed by the CSI driver during NodeUnpublishVolume", volpath)
+		})
 	})
 
 	Describe("NodeStageVolume", func() {
--- a/pkg/sanity/sanity.go
+++ b/pkg/sanity/sanity.go
@@ -164,6 +164,14 @@ type TestConfig struct {
 	// n > 0: repeat each call n times
 	// NewTestConfig() by default enables idempotency testing.
 	IdempotentCount int
+
+	// CheckPath is a callback function to check whether the given path exists.
+	// If this is not set, then defaultCheckPath will be used instead.
+	CheckPath func(path string) (PathKind, error)
+	// Command to be executed for a customized way to check a given path.
+	CheckPathCmd string
+	// Timeout for the executed command to check a given path.
+	CheckPathCmdTimeout time.Duration
 }
 
 // TestContext gets initialized by the sanity package before each test
@@ -194,6 +202,7 @@ func NewTestConfig() TestConfig {
 		TestVolumeAccessType: "mount",
 		IDGen:                &DefaultIDGenerator{},
 		IdempotentCount:      10,
+		CheckPathCmdTimeout:  10 * time.Second,
 
 		DialOptions:           []grpc.DialOption{grpc.WithInsecure()},
 		ControllerDialOptions: []grpc.DialOption{grpc.WithInsecure()},
@@ -448,4 +457,90 @@ func PseudoUUID() string {
 // alone should already be fairly unique.
 func UniqueString(prefix string) string {
 	return prefix + uniqueSuffix
+}
+
+// Return codes for CheckPath
+type PathKind string
+
+const (
+	PathIsFile     PathKind = "file"
+	PathIsDir      PathKind = "directory"
+	PathIsNotFound PathKind = "not_found"
+	PathIsOther    PathKind = "other"
+)
+
+// IsPathKind validates that the input string matches one of the defined
+// PathKind values above. If successful, it returns the corresponding
+// PathKind type. Otherwise, it returns an error.
+func IsPathKind(in string) (PathKind, error) {
+	pk := PathKind(in)
+	switch pk {
+	case PathIsFile, PathIsDir, PathIsNotFound, PathIsOther:
+		return pk, nil
+	default:
+		return pk, fmt.Errorf("invalid PathType: %s", pk)
+	}
+}
+
+// defaultCheckPath runs os.Stat against the provided path and returns
+// a code indicating whether it's a file, directory, not found, or other.
+// If an error occurs, it returns an empty string along with the error.
+func defaultCheckPath(path string) (PathKind, error) {
+	var pk PathKind
+	fi, err := os.Stat(path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return PathIsNotFound, nil
+		} else {
+			return "", err
+		}
+	}
+	switch mode := fi.Mode(); {
+	case mode.IsRegular():
+		pk = PathIsFile
+	case mode.IsDir():
+		pk = PathIsDir
+	default:
+		pk = PathIsOther
+	}
+	return pk, nil
+}
+
+// CheckPath takes a path parameter and returns a code indicating whether
+// it's a file, directory, not found, or other. This can be done using a
+// custom command, custom function, or by the defaultCheckPath function.
+// If an error occurs, it returns an empty string along with the error.
+func CheckPath(path string, config *TestConfig) (PathKind, error) {
+	if path == "" {
+		return "", fmt.Errorf("path argument must not be empty")
+	}
+	if config == nil {
+		return "", fmt.Errorf("config argument must not be nil")
+	}
+
+	if config.CheckPathCmd != "" {
+		// Check the provided path using the check path command.
+		ctx, cancel := context.WithTimeout(context.Background(), config.CheckPathCmdTimeout)
+		defer cancel()
+
+		cmd := exec.CommandContext(ctx, config.CheckPathCmd, path)
+		cmd.Stderr = os.Stderr
+		out, err := cmd.Output()
+		if err != nil {
+			return "", fmt.Errorf("check path command %s failed: %v", config.CheckPathCmd, err)
+		}
+		// The output of this command is expected to match the value for
+		// PathIsFile, PathIsDir, PathIsNotFound, or PathIsOther.
+		pk, err := IsPathKind(strings.TrimSpace(string(out)))
+		if err != nil {
+			return "", fmt.Errorf("check path command %s failed: %v", config.CheckPathCmd, err)
+		}
+		return pk, nil
+	} else if config.CheckPath != nil {
+		// Check the path using a custom callback function.
+		return config.CheckPath(path)
+	} else {
+		// Use defaultCheckPath if no custom function was provided.
+		return defaultCheckPath(path)
+	}
 }
--- a/hack/e2e.sh
+++ b/hack/e2e.sh
@@ -101,9 +101,23 @@ echo $targetpath
 rm -rf $@
 ' > custompathremoval.bash
 
-	chmod +x custompathcreation.bash custompathremoval.bash
+	# Create a script for custom target path check.
+	echo '#!/bin/bash
+if [ -f "$1" ]; then
+    echo "file"
+elif [ -d "$1" ]; then
+    echo "directory"
+elif [ -e "$1" ]; then
+    echo "other"
+else
+    echo "not_found"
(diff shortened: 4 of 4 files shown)


pkg/sanity/node.go
```diff
@@ -449,6 +449,67 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
...
+				&csi.NodeUnpublishVolumeRequest{
+					VolumeId:   volid,
+					TargetPath: volpath,
+				},
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			// NodeGetVolumeStats is expected to return NotFound
+			// if the TargetPath is removed by NodeUnpublishVolume
+			By("Checking NodeGetVolumeStats returns NotFound error")
```
pohly, line 500: This seems to make assumptions about how `NodeGetVolumeStats` is implemented. It could return an error because it tracks volume state and knows that `NodeUnpublishVolume` was called even when that function didn't remove the target path.
 
What you really need is to check on the node whether the directory is gone. Perhaps you can use `CreateTargetDir` (https://github.com/kubernetes-csi/csi-test/blob/a251c44fd49d9eedd55a40b71e5da4ad080ba431/pkg/sanity/sanity.go#L124) or whatever is configured (https://github.com/kubernetes-csi/csi-test/blob/a251c44fd49d9eedd55a40b71e5da4ad080ba431/pkg/sanity/sanity.go#L143). At least the internal implementation (https://github.com/kubernetes-csi/csi-test/blob/a251c44fd49d9eedd55a40b71e5da4ad080ba431/pkg/sanity/sanity.go#L362-L367) fails when the directory exists.

But that looks like a hack. A cleaner solution is to introduce a `CheckPath` feature (with a configurable callback and command) which returns "file", "directory", "other" and "not_found", or an error, and then use that here.

dobsonj: Thanks for the feedback @pohly, this was very helpful.
I implemented the CheckPath function, but with a couple of minor differences from what you described.
1) I used return codes instead of strings, but it does return the 4 conditions you described. The test in node.go only checks for found or not found though. Since the spec states "file or directory" I did not want to make assumptions about the file type in the sanity test.
2) I only did the configurable callback, not the command string that can override that behavior like in createMountTargetLocation() and removeMountTargetLocation(). This simplified the implementation, and I question whether the command string would be as useful in this context as it is for create / remove commands. Translating those 4 conditions into error codes (or strings) would add some complexity to the command. If the command string supports an essential use case though, I can add in.

pohly: The command string is essential for those users who invoke the `csi-sanity` command to run tests. They can't extend that binary with Go callbacks.

dobsonj: Thanks for elaborating on this, I added the command string option.
I tested it with the following `check_path.sh` bash script:
```bash
#!/bin/bash
if [ -f "$1" ]; then
        echo "1" # file
elif [ -d "$1" ]; then
        echo "2" # dir
elif [ ! -e "$1" ]; then
        echo "3" # not found
else
        echo "4" # other
fi
```
And that passes against the mock driver using the following args:
```
$ csi-sanity -csi.checkpathcmd ./check_path.sh -csi.endpoint /tmp/csi.sock
```

pohly: Why use numbers as output? I find strings as in my original proposal easier to use because there's no chance that someone, say, returns 3 when they meant 4. The same strings can be string constants in Go for the Go callback, so it could be even made type safe there.

dobsonj: You're right, strings are a better approach here. When I added the `Atoi` call I started to question if that was really the best option. Updated.

pkg/sanity/sanity.go
```diff
@@ -449,3 +460,68 @@ func PseudoUUID() string {
...
+		defer cancel()
+
+		cmd := exec.CommandContext(ctx, config.CheckPathCmd, path)
+		cmd.Stderr = os.Stderr
+		out, err := cmd.Output()
+		if err != nil {
+			return "", fmt.Errorf("check path command %s failed: %v", config.CheckPathCmd, err)
+		}
+		// Convert the command's stdout to an integer. This is expected to match
+		// the value for PathIsFile, PathIsDir, PathIsNotFound, or PathIsOther.
```
pohly, line 520: This comment is out-dated, right?

pkg/sanity/sanity.go
```diff
@@ -449,3 +460,68 @@ func PseudoUUID() string {
 func UniqueString(prefix string) string {
 	return prefix + uniqueSuffix
 }
+
+// Return codes for CheckPath
```
pohly, line 464: Please introduce a type (like `type PathKind string`) and use that to make the code more type safe.

pkg/sanity/sanity.go
```diff
@@ -449,3 +460,68 @@ func PseudoUUID() string {
...
+
+		cmd := exec.CommandContext(ctx, config.CheckPathCmd, path)
+		cmd.Stderr = os.Stderr
+		out, err := cmd.Output()
+		if err != nil {
+			return "", fmt.Errorf("check path command %s failed: %v", config.CheckPathCmd, err)
+		}
+		// Convert the command's stdout to an integer. This is expected to match
+		// the value for PathIsFile, PathIsDir, PathIsNotFound, or PathIsOther.
+		outstr := strings.TrimSpace(string(out))
```
pohly, line 521: Here it would be good to check that the script has return a valid string. If not, raise an error.

pohly (changes requested):

pkg/sanity/sanity.go
dobsonj: Yes, good catch.

pkg/sanity/sanity.go
dobsonj: Done

pkg/sanity/sanity.go
dobsonj: Done

pkg/sanity/sanity.go
```diff
@@ -449,3 +459,86 @@ func PseudoUUID() string {
...
+	var err error = nil
+	switch pk {
+	case PathIsFile:
+	case PathIsDir:
+	case PathIsNotFound:
+	case PathIsOther:
+	default:
+		err = fmt.Errorf("invalid PathType: %s", pk)
+	}
+	return err
```
pohly, line 483: This seems non-idiomatic. A shorter version is:
```
func (pk PathKind) validate() error {
	switch pk {
	case PathIsFile, PathIsDir, PathIsNotFound, PathIsOther:
              return nil
         default:
		return fmt.Errorf("invalid PathType: %s", pk)
	}
}
```

pohly: I also find it odd to first cast a string and then call validate. It works of course, but a `func IsPathKind (in string) (PathKind, error)` makes the intent clearer.

pohly (changes requested):
Please squash your commits.

Except for some nits it looks okay, but I want to try it out before merging. I should get to that on Monday.

pkg/sanity/sanity.go
dobsonj: Makes sense, done. Appreciate your help with this.

cmd/csi-sanity/main.go
```diff
@@ -81,6 +81,8 @@ func main() {
 	stringVar(&config.RemoveTargetPathCmd, "removemountpathcmd", "Command to run for target path removal")
 	stringVar(&config.RemoveStagingPathCmd, "removestagingpathcmd", "Command to run for staging path removal")
 	durationVar(&config.RemovePathCmdTimeout, "removepathcmdtimeout", "Timeout for the commands to remove target and staging paths, in seconds")
+	stringVar(&config.CheckPathCmd, "checkpathcmd", "Command to run to check a given path")
```
pohly, line 84: It's a bit hard for users of csi-sanity to find out what that command needs to do. Right now they have to dig into the source code to find the command in `pkg/sanity/sanity.go`.

Perhaps add `It must print "file", "directory", "not_found", "other" on stdout.`?

pkg/sanity/node.go
```diff
@@ -449,6 +449,51 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
 			Expect(ok).To(BeTrue())
 			Expect(serverError.Code()).To(Equal(codes.InvalidArgument), "unexpected error: %s", serverError.Message())
 		})
+
+		It("should fail if target path was not removed by driver", func() {
```
pohly, line 453: Please replace with "should remove target path". The full test name then becomes `NodeUnpublishVolume should remove target path", which is a valid statement about the behavior of the driver.

pohly (commented):
Almost done. I tried this out in PMEM-CSI (https://github.com/intel/pmem-csi/pull/948) and it worked as expected.

cmd/csi-sanity/main.go
dobsonj: Done

pkg/sanity/node.go
dobsonj: Done

gnufied:
Once this merges - we should send out an email to CSI announce list on - https://github.com/container-storage-interface/community so as CSI driver authors can fix their drivers.

pohly:
> Once this merges - we should send out an email to CSI announce list on - https://github.com/container-storage-interface/community so as CSI driver authors can fix their drivers.

That change of behavior in Kubernetes was already announced in https://groups.google.com/g/container-storage-interface-drivers-announce/c/GgtP9kiv5Qc/m/CO2yyOJvAAAJ

That announcement did not specifically talk about deleting the directory, though, because it follows from the spec. Pointing out the change of behavior of the csi-sanity suite is still worthwhile to avoid surprises.

pohly:
Is the mock driver not creating the target path? That might have to be added for pull-kubernetes-csi-csi-test to pass.

pohly:
That the test checks the path when it should exist is useful because it informs users of csi-sanity when their check path configuration isn't working. Please don't remove that.

dobsonj:
> Is the mock driver not creating the target path? That might have to be added for pull-kubernetes-csi-csi-test to pass.

The mock driver works okay, but `pull-kubernetes-csi-csi-test` uses hostpath. I'm looking into that.

pohly:
> > Is the mock driver not creating the target path? That might have to be added for pull-kubernetes-csi-csi-test to pass.
> 
> The mock driver works okay, but `pull-kubernetes-csi-csi-test` uses hostpath. I'm looking into that.

Then this might be exactly the issue that I mentioned: the check path feature must be configured so that it checks the path where hostpath is deployed, not where the csi-sanity test runs.

The test assertion failure could be enhanced. I'll leave a comment on the code line.

pkg/sanity/node.go
```diff
@@ -449,6 +449,51 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
...
+			conpubvol := controllerPublishVolume(name, vol, nid)
+			_ = nodeStageVolume(name, vol, conpubvol)
+			_ = nodePublishVolume(name, vol, conpubvol)
+
+			// Verify that the path exists before calling
+			// NodeUnpublishVolume.
+			By("Checking the target path exists")
+			pa, err := CheckPath(volpath, sc.Config)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(pa).NotTo(Equal(PathIsNotFound))
```
pohly, line 477: When this assertion fails, the resulting errors is not very helpful:
```
/home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:453
Expected
    <sanity.PathKind>: not_found
not to equal
    <sanity.PathKind>: not_found
/home/prow/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/node.go:477
```

You can add a description to it:
```
Expect(err).NotTo(HaveOccurred(), "checking path %q", volpath)
Expect(pa).NotTo(Equal(PathIsNotFound), "path %q should have been created by CSI driver and the test config should enabling testing for that path", volpath)
```

pkg/sanity/node.go
```diff
@@ -449,6 +449,51 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
...
+			)
+			Expect(err).NotTo(HaveOccurred())
+
+			// The CSI spec states that the SP MUST delete
+			// the file or directory it created at this path
+			// as part of NodeUnpublishVolume.
+			By("Checking the target path was removed")
+			pa, err = CheckPath(volpath, sc.Config)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(pa).To(Equal(PathIsNotFound))
```
pohly, line 495: Similar here.

dobsonj:
> Then this might be exactly the issue that I mentioned: the check path feature must be configured so that it checks the path where hostpath is deployed, not where the csi-sanity test runs.

Yeah, now I see, it needs to be configured here:
https://github.com/kubernetes-csi/csi-test/blob/a251c44fd49d9eedd55a40b71e5da4ad080ba431/release-tools/prow.sh#L976-L985
I'll follow up with a fix for this and the other 2 comments about improving the error messages.

pohly:
> I'll follow up with a fix for this and the other 2 comments about improving the error messages.

`release-tools` cannot be modified here. Instead, `csi-release-tools` has to be modified and then gets imported into various projects.

This failure shows that the new test is a breaking change: users of csi-sanity have to do something when updating, otherwise they get test failures. In this repo that causes a circular dependency: we first have to update csi-release-tools, but that then only works with an update csi-test.

Let's avoid this breaking change. Two options:
- Don't set any default for the check path feature: if unset, skip the test.
- If creating the path is not the default, expect path checking also to be reconfigured, otherwise skip the test.

The first options looks simpler at first glance, but then the `csi-sanity` command needs a parameter for setting the builtin default implementation. This could be hard to explain.

Therefore I prefer the second option.

dobsonj:
> `release-tools` cannot be modified here. Instead, `csi-release-tools` has to be modified and then gets imported into various projects.
> 
> This failure shows that the new test is a breaking change: users of csi-sanity have to do something when updating, otherwise they get test failures. In this repo that causes a circular dependency: we first have to update csi-release-tools, but that then only works with an update csi-test.
> 
> Let's avoid this breaking change. Two options:
> 
> * Don't set any default for the check path feature: if unset, skip the test.
> * If creating the path is not the default, expect path checking also to be reconfigured, otherwise skip the test.
> 
> The first options looks simpler at first glance, but then the `csi-sanity` command needs a parameter for setting the builtin default implementation. This could be hard to explain.
> 
> Therefore I prefer the second option.

This makes sense, I'll go with the second option as you suggest.

I noticed after I pushed those changes that `make test` started failing the `verify-subtree.sh` check with `Directory 'release-tools' contains non-upstream changes`. One thing that still puzzles me though is why `csi-sanity` still failed on the same line after that last change. Could be an issue with the change that I just pushed to prow.sh, or an issue with the hostpath driver (though I didn't notice an obvious problem when reading the code). But is it possible that my change to the local copy of `release-tools` was somehow ignored during the test? Just trying to better understand the full process.

In any case, I'll revert the change to `release-tools` and go with option 2 for this current PR.

pohly:
> But is it possible that my change to the local copy of `release-tools` was somehow ignored during the test?

No, I see `csi-sanity` being called with `-csi.checkpathcmd /home/prow/go/pkg/csiprow.ZLUhMcmtDc/checkdir_in_pod.sh`.

release-tools/prow.sh
```diff
@@ -968,6 +968,21 @@ if ! kubectl exec "${CSI_PROW_SANITY_POD}" -c "${CSI_PROW_SANITY_CONTAINER}" --
     exit 1
 fi
 EOF
+
+    cat >"${CSI_PROW_WORK}/checkdir_in_pod.sh" <<EOF
+#!/bin/sh
+OUT=\$(kubectl exec "${CSI_PROW_SANITY_POD}" -c "${CSI_PROW_SANITY_CONTAINER}" -- stat --printf="%F" "\$@" 2>&1)
+if [ \$? -ne 0 ]; then
```
pohly, line 975: I don't see any obvious issue with this, but treating *any* error (including failures in kubectl) as "not_found" doesn't look right.

For example, does `stat` in the container support `--printf`?

I briefly considered using `stat`, but then decided against as too brittle. Instead I went with a sequence of standard shell if checks in https://github.com/intel/pmem-csi/pull/948/files#diff-67bdd13daabc0c2eb2feaba824f0b9668e60c4cd9a68cad31357eaa568c50e8a

dobsonj: I see, thank you, that seems the most likely explanation for the failure in the last test run (i.e. returning non-zero for some unexpected error).

dobsonj:
I dropped the commit that changed `release-tools/prow.sh`, made minor changes to the error messages in `node.go`, and then implemented the skip test check as we discussed. This change also exposed a place in `hack/e2e.sh` that needed to be updated to use the new `--csi.checkpathcmd` argument.

dobsonj:
CI passed this time, I'll wait to squash the commits again until the review is otherwise complete.

pkg/sanity/node.go
```diff
@@ -451,6 +451,19 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
...
+			// custom target path functions if they have not yet
+			// implemented similar functionality to check if the
+			// path exists. Skip this test if there is a custom
+			// command or function provided to create the path,
+			// but not yet provided to check the path.
+			if sc.Config.CreateTargetPathCmd != "" && sc.Config.CheckPathCmd == "" {
+				Skip("CreateTargetPathCmd was set, but CheckPathCmd was not. Please update your testing configuration to enable CheckPathCmd.")
+			}
+			if sc.Config.CreateTargetDir != nil && sc.Config.CheckPath == nil {
+				Skip("CreateTargetDir was set, but CheckPath was not. Please update your testing configuration to enable CheckPath.")
```
pohly, line 464: Both should always be non-nil because they have defaults.

What you need to check for here is "sc.Config.CreateTargetDir not default and sc.Config.CheckPath not default".

Also, I think this only needs to be checked when `sc.Config.CreateTargetPathCmd` is unset. Otherwise we check and use `sc.Config.CheckPathCmd`.

dobsonj: `createMountTargetLocation` has 2 separate paths, one for the custom function, but if that is nil then it calls mkdir. `sc.Config.CreateTargetDir` is nil by default.
https://github.com/dobsonj/csi-test/blob/a2091104af4d4e1df72bc6cfa8de4d955afa411a/pkg/sanity/sanity.go#L363-L378
Originally I did try to compare `sc.Config.CheckPath != defaultCheckPath` here, but got the following compile error:
```
pkg/sanity/node.go:463:63: invalid operation: sc.Config.CheckPath != defaultCheckPath (func can only be compared to nil)
```
Which is interesting... I think that should work fine in C, but apparently not in Go.
So I ended up changing `CheckPath` to follow the same model as `createMountTargetLocation` and leave `sc.Config.CheckPath` set to nil by default, specifically so the check on line 463 would work.
https://github.com/dobsonj/csi-test/blob/a2091104af4d4e1df72bc6cfa8de4d955afa411a/pkg/sanity/sanity.go#L539-L545

If `sc.Config.CreateTargetPathCmd` is set, we should hit the skip statement on line 461 and not make it down to line 464. Unless `CreateTargetPathCmd`, `CheckPathCmd`, and `CreateTargetDir` are all set at the same time, which seems strange. I can add a check for that though if that's what you're pointing out.

pohly: > So I ended up changing CheckPath to follow the same model as createMountTargetLocation and leave sc.Config.CheckPath set to nil by default, specifically so the check on line 463 would work.

That's the part that I had missed. Then the current code is fine.

pohly (changes requested):
Looks good now, but please squash before we merge it.

dobsonj:
> Looks good now, but please squash before we merge it.

Done, thanks.

pohly (commented):
/lgtm
/approve

PR #349, opened 2021-08-19
support node VOLUME_MOUNT_GROUP cap in test

andyzhangx:
**What type of PR is this?**
/kind failing-test

support node VOLUME_MOUNT_GROUP cap, this is a new node capability added in CSI v1.5.0 spec

Fixes #

```release-note
support node VOLUME_MOUNT_GROUP cap in test
```

### DIFF
--- a/pkg/sanity/node.go
+++ b/pkg/sanity/node.go
@@ -342,6 +342,7 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
 				case csi.NodeServiceCapability_RPC_EXPAND_VOLUME:
 				case csi.NodeServiceCapability_RPC_VOLUME_CONDITION:
 				case csi.NodeServiceCapability_RPC_SINGLE_NODE_MULTI_WRITER:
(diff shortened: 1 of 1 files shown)


pohly:
Does this change affect testing CSI drivers in any way?

gnufied:
@pohly yep you are exactly right. Without this change, I would assume the drivers that advertise this capability will fail the sanity test. It does have a release-note doesn't it?

pohly:
> It does have a release-note doesn't it?

Yes, but it wasn't obvious what  "support node VOLUME_MOUNT_GROUP cap in test" means (who needs this support, under which circumstances?).

Can we change it into "avoid test failures when a CSI driver advertises the VOLUME_MOUNT_GROUP capability from CSI 1.5"?

> shall we have a new release for this change?

That is what we were discussing yesterday in the CSI meeting. Because the purpose of this PR wasn't clear, we didn't come to a conclusion. But now it seems that we should do a new release because of this issue.

PR #363, opened 2022-03-13
add ControllerPublishSupported

### LINKED ISSUE #316: LifeCycle test fails since 'unpublish' is not run
On branch master - in "volume lifecycle" test (func VolumeLifecycle), seems like the last method running is ControllerPublishVolumeRequest and nothing after that, which causes the test to fail since it expects unpublish.

on the other hand, on branch release-3.1, Lifecycle tests pass.

This is the failure from master branch:

```
CSI Driver Test Suite.Controller Service [Controller Server] volume lifecycle should workFailing for the past 2 builds (Since #580 )Took 0.44 sec.add descriptionStacktrace/usr/local/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:119
recorded 1 failure(s)
/go/pkg/mod/github.com/onsi/ginkgo@v1.10.3/internal/leafnodes/runner.go:113Standard Output�[1mSTEP�[0m: reusing connection to CSI driver at /tmp/k8s_dir/nodecsi
�[1mSTEP�[0m: reusing connection to CSI driver controller at /tmp/k8s_dir/f
�[1mSTEP�[0m: creating mount and staging directories
�[1mSTEP�[0m: getting node information
�[1mSTEP�[0m: creating a single node writer volume
�[1mSTEP�[0m: calling controllerpublish on that volume
cleanup volumes: deleting A9000:6001738CFC9035EA000000000078E5C7
cleanup volumes: DeleteVolume failed: rpc error: code = Internal desc = an internal exception occurred : The volume mapped to a host cannot be deleted.

and:

CSI Driver Test Suite.Controller Service [Controller Server] volume lifecycle should be idempotentFailing for the past 2 builds (Since #580 )Took 23 ms.add descriptionStacktrace/usr/local/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:1066
volume create failed
Unexpected error:
    <*status.Error | 0xc0001fe550>: {
        Code: 6,
        Message: "Volume was already created with different size.",
        Details: nil,
        XXX_NoUnkeyedLiteral: {},
        XXX_unrecognized: nil,
        XXX_sizecache: 0,
    }
    rpc error: code = AlreadyExists desc = Volume was already created with different size.
occurred
/usr/local/go/src/github.com/kubernetes-csi/csi-test/pkg/sanity/controller.go:1659Standard Output�[1mSTEP�[0m: reusing connection to CSI driver at /tmp/k8s_dir/nodecsi
�[1mSTEP�[0m: reusing connection to CSI driver controller at /tmp/k8s_dir/f
�[1mSTEP�[0m: creating mount and staging directories
�[1mSTEP�[0m: getting node information
�[1mSTEP�[0m: creating a single node writer volume
```

ArbelNathan:
**What type of PR is this?**

/kind failing-test

The lack of `ControllerPublishSupported` in `Resources` cause [this line](https://github.com/kubernetes-csi/csi-test/blob/da7a71c90b602694a44815fc4c3533e803601b87/pkg/sanity/resources.go#L295) to always return false.

Fixes #316

if you see a more elegant way to enter the `ControllerPublishSupported` into `Resources` please feel free to fix it.

```release-note
None
```

### DIFF
--- a/pkg/sanity/node.go
+++ b/pkg/sanity/node.go
@@ -310,11 +310,9 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
 		nodeExpansionSupported = isNodeCapabilitySupported(n, csi.NodeServiceCapability_RPC_EXPAND_VOLUME)
 		controllerExpansionSupported = isControllerCapabilitySupported(cl, csi.ControllerServiceCapability_RPC_EXPAND_VOLUME)
 		r = &Resources{
-			Context:                    sc,
-			ControllerClient:           cl,
-			NodeClient:                 n,
-			ControllerPublishSupported: controllerPublishSupported,
-			NodeStageSupported:         nodeStageSupported,
+			Context:          sc,
+			ControllerClient: cl,
+			NodeClient:       n,
 		}
 	})
 
--- a/pkg/sanity/resources.go
+++ b/pkg/sanity/resources.go
@@ -65,8 +65,6 @@ type Resources struct {
 	// directly if automatic cleanup is not desired and cannot be avoided
 	// otherwise.
 	csi.NodeClient
-	ControllerPublishSupported bool
-	NodeStageSupported         bool
 
 	// mutex protects access to managedResourceInfos.
 	mutex                sync.Mutex
@@ -279,7 +277,7 @@ func (cl *Resources) cleanupVolume(ctx context.Context, offset int, volumeID str
 			errs = append(errs, fmt.Errorf("NodeUnpublishVolume for volume ID %s failed: %s", volumeID, err))
 		}
 
-		if cl.NodeStageSupported {
+		if isNodeCapabilitySupported(cl, csi.NodeServiceCapability_RPC_STAGE_UNSTAGE_VOLUME) {
 			if _, err := cl.NodeUnstageVolume(
 				ctx,
 				&csi.NodeUnstageVolumeRequest{
@@ -292,7 +290,7 @@ func (cl *Resources) cleanupVolume(ctx context.Context, offset int, volumeID str
 		}
 	}
 
-	if cl.ControllerPublishSupported && info.NodeID != "" {
+	if info.NodeID != "" {
 		if _, err := cl.ControllerClient.ControllerUnpublishVolume(
 			ctx,
 			&csi.ControllerUnpublishVolumeRequest{


pkg/sanity/controller.go
```diff
@@ -114,6 +114,7 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 			ControllerClient: csi.NewControllerClient(sc.ControllerConn),
 			NodeClient:       csi.NewNodeClient(sc.Conn),
 		}
+		r.ControllerPublishSupported = isControllerCapabilitySupported(r, csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME)
```
pohly, line 117: If `r.ControllerPublishSupported` was always false (= unset), why was `ControllerPublishVolume` called (as mentioned in https://github.com/kubernetes-csi/csi-test/issues/316)?

Are there perhaps `ControllerPublishVolume` calls which need to be guarded with an if check?

ArbelNathan: Every test that uses `ControllerPublishVolume` has its [own guard](https://github.com/kubernetes-csi/csi-test/blob/b486fd25346ea68b3c365e96fb7241df8adf64dd/pkg/sanity/controller.go#L1059-L1062)..

If only `ControllerPublishVolume` [can set the](https://github.com/kubernetes-csi/csi-test/blob/b486fd25346ea68b3c365e96fb7241df8adf64dd/pkg/sanity/resources.go#L151) `volumeInfo`'s `NodeId` so maybe [that cl.ControllerPublishSupported check](https://github.com/kubernetes-csi/csi-test/blob/da7a71c90b602694a44815fc4c3533e803601b87/pkg/sanity/resources.go#L295) is redundant and we should just leave the `info.NodeID != ""` there (and remove `ControllerPublishSupported` in `Resources` I added), but I don't know If we wanna just removed guards.

pohly: I prefer to remove `ControllerPublishSupported`. The code becomes simpler without it and we don't need it because when `info.NodeID != ""`, then we know that `ControllerUnpublishVolume` can and must be called.

pohly: Let's also remove `NodeStageSupported`. It's only checked once, which can replaced with a `isNodeCapabilitySupported` call.

ArbelNathan: ok, done.

ArbelNathan:
> This looks good, thanks for investigating. Can you squash into a single commit?

squashed

pohly (approved):
/lgtm
/ok-to-test

PR #366, opened 2022-04-19
ginkgo v2

pohly:
**What type of PR is this?**
/kind cleanup

Ginkgo v1 is deprecated and was replaced with the v2 rewrite.

```release-note
csi-sanity is now built with ginkgo v2. All camel case flags (e.g. -randomizeAllSpecs) are replaced with kebab case flags (e.g. -randomize-all-specs). The camel case versions continue to work but emit a deprecation warning. The "junitfile" configuration file option was removed because ginkgo now handles JUnit report creation itself.
```


pohly:
We can probably merge it at the same time as the Kubernetes switch. I want to have that ready first to avoid a situation where we switch in this repo, but then Kubernetes doesn't.

VirrageS:
@pohly From what I see, Kubernetes MR to move to Ginkgo V2 has been merged. Could we merge this as well?

pohly:
@VirrageS: I've updated the PR and think that we can merge it now.

I'm undecided whether I should treat this as a breaking change (thus leading to csi-test v2.0.0). I'm leaning towards not bumping the major version because it doesn't really help with the migration (in contrast to other Go module version changes, ginkgo v1 and v2 cannot co-exist in the same binary) and the difference for command line users might not matter.

pohly:
@VirrageS: any thoughts on this? Can you perhaps review?

VirrageS:
I've look at the changes, LGTM.

> I'm undecided whether I should treat this as a breaking change (thus leading to csi-test v2.0.0). I'm leaning towards not bumping the major version because it doesn't really help with the migration (in contrast to other Go module version changes, ginkgo v1 and v2 cannot co-exist in the same binary) and the difference for command line users might not matter.

I think `csi-test` is currently at `v4` actually. I would say that this is quite breaking change as I think people who are using `csi-test` right now with Ginkgo v1 in their codebase might stumble upon incompatibilities. I think we should bump `csi-test` to `v5`. Wdyt?

pohly:
Okay, let's bump to v5 then.

PR #450, opened 2023-04-25
Allow drivers to support the GROUP_CONTROLLER_SERVICE capability

nixpanic:
**What type of PR is this?**

/kind feature

The external-snapshotter is going to support [Volume Group Snapshot](kubernetes/enhancements#1551), so drivers will return the capability as well.

Fixes #

There are no tests for the new Volume Group Snapshot yet. kubernetes-csi/csi-driver-host-path#399 is the first PR for a driver to support the new ALPHA feature.

```release-note
NONE
```


xing-yang:
Can you add a release note?

PR #476, opened 2023-08-08
always use unique name for volume and snapshot

huww98:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Always add random chars to name.

This would avoid any name conflicts. And matches k8s behaviour better.

Some system does not work if the same name is reused even after the previous resource is deleted. The idempotent state may not be cleared when resource deleted.

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/sanity/sanity.go
+++ b/pkg/sanity/sanity.go
@@ -453,6 +453,14 @@ func UniqueString(prefix string) string {
 	return prefix + uniqueSuffix
 }
 
+func UniqueStringWithLength(prefix string, length int) string {
+	str := UniqueString(prefix)
+	if len(str) > length {
+		panic(fmt.Sprintf("prefix %q is too long, use a shorter one", prefix))
+	}
+	return str + strings.Repeat("a", length-len(str))
+}
+
 // Return codes for CheckPath
 type PathKind string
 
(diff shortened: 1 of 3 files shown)


huww98:
I think this PR is still relevant. @msau42 @gnufied @RaunakShah PTAL~.

pkg/sanity/sanity.go
```diff
@@ -453,6 +453,14 @@ func UniqueString(prefix string) string {
 	return prefix + uniqueSuffix
 }
 
+func UniqueStringWithLength(prefix string, length int) string {
+	str := UniqueString(prefix)
+	if len(str) > length {
+		Panic()
```
xing-yang, line 459: Do we have to introduce Panic() here?  Can we trim it if it is longer than "length" and make it a best effort to have a unique name?

huww98: I think we'd better panic here, and make it guaranteed unique. All callers are just using a hard-coded prefix, So if someone do reach here, it should be easy for him to just fix the panic by using a shorter prefix, rather than debugging weird issue caused by non-unique ID.

I mean panic() rather than Panic() here, I'll fix that.

PR #523, opened 2024-04-02
replace offset args with new GinkgoHelper

huww98:
**What type of PR is this?**
>
> /kind api-change
> /kind bug

/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

it is simpler, and easier to maintain

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/sanity/resources.go
+++ b/pkg/sanity/resources.go
@@ -77,63 +77,36 @@ type Resources struct {
 
 // ControllerClient interface wrappers
 
-// CreateVolume proxies to a Controller service implementation and registers the
-// volume for cleanup.
-func (cl *Resources) CreateVolume(ctx context.Context, in *csi.CreateVolumeRequest, _ ...grpc.CallOption) (*csi.CreateVolumeResponse, error) {
-	return cl.createVolume(ctx, 2, in)
-}
-
-// DeleteVolume proxies to a Controller service implementation and unregisters
-// the volume from cleanup.
-func (cl *Resources) DeleteVolume(ctx context.Context, in *csi.DeleteVolumeRequest, _ ...grpc.CallOption) (*csi.DeleteVolumeResponse, error) {
-	return cl.deleteVolume(ctx, 2, in)
-}
-
-// ControllerPublishVolume proxies to a Controller service implementation and
-// adds the node ID to the corresponding volume for cleanup.
-func (cl *Resources) ControllerPublishVolume(ctx context.Context, in *csi.ControllerPublishVolumeRequest, _ ...grpc.CallOption) (*csi.ControllerPublishVolumeResponse, error) {
-	return cl.controllerPublishVolume(ctx, 2, in)
-}
-
-// CreateSnapshot proxies to a Controller service implementation and registers
-// the snapshot for cleanup.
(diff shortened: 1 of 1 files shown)


pkg/sanity/resources.go
```diff
@@ -77,88 +77,65 @@ type Resources struct {
...
 	}
 	return conpubvol, err
 }
 
 // registerVolume adds or updates an entry for given volume.
-func (cl *Resources) registerVolume(offset int, id string, info volumeInfo) {
-	ExpectWithOffset(offset, id).NotTo(BeEmpty(), "volume ID is empty")
-	ExpectWithOffset(offset, info).NotTo(BeNil(), "volume info is nil")
+func (cl *Resources) registerVolume(id string, info volumeInfo) {
+	Expect(id).NotTo(BeEmpty(), "volume ID is empty")
```
pohly, line 137: Missing `ginkgo.GinkgoHelper` here.

pkg/sanity/resources.go
```diff
@@ -171,65 +148,69 @@ func (cl *Resources) registerVolume(offset int, id string, info volumeInfo) {
...
 	cl.mutex.Lock()
 	defer cl.mutex.Unlock()
-	cl.registerSnapshotNoLock(offset+1, id)
+	cl.registerSnapshotNoLock(id)
 }
 
-func (cl *Resources) registerSnapshotNoLock(offset int, id string) {
-	ExpectWithOffset(offset, id).NotTo(BeEmpty(), "ID for register snapshot is missing")
+func (cl *Resources) registerSnapshotNoLock(id string) {
+	Expect(id).NotTo(BeEmpty(), "ID for register snapshot is missing")
```
pohly, line 198: Missing `ginkgo.GinkgoHelper` here.

pkg/sanity/resources.go
```diff
@@ -171,65 +148,69 @@ func (cl *Resources) registerVolume(offset int, id string, info volumeInfo) {
...
 	cl.mutex.Lock()
 	defer cl.mutex.Unlock()
-	cl.unregisterResourceNoLock(offset+1, id)
+	cl.unregisterResourceNoLock(id)
 }
 
-func (cl *Resources) unregisterResourceNoLock(offset int, id string) {
-	ExpectWithOffset(offset, id).NotTo(BeEmpty(), "ID for unregister resource is missing")
+func (cl *Resources) unregisterResourceNoLock(id string) {
+	Expect(id).NotTo(BeEmpty(), "ID for unregister resource is missing")
```
pohly, line 213: Missing `ginkgo.GinkgoHelper` here.

pohly (changes requested):
`ginkgo.GinkgoHelper` didn't exist when this code was written. Using it indeed is simpler.

We want to treat all methods as helpers, right? Then some `ginkgo.GinkgoHelper` calls are missing.

huww98:
> We want to treat all methods as helpers, right?

Yes, but so long as one of the callers has `GinkgoHelper()`, it should be fine. And I think `registerXxx()` is not intended to be called outside of resources.go.

pohly:
> Yes, but so long as one of the callers has GinkgoHelper(), it should be fine.

That's not my understanding of how that function works. [GinkgoHelper](https://pkg.go.dev/github.com/onsi/ginkgo/v2#GinkgoHelper) marks "*the function* it's called in as a test helper" (emphasis mine). It does not mark the function and all functions called by it as helper.

PR #524, opened 2024-04-02
use MakeCreateVolumeReq()

huww98:
**What type of PR is this?**
>
> /kind api-change

/kind bug
/kind cleanup

> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Use MakeCreateVolumeReq for all tests that does not care how volumes are created.

This fixes several places where TestVolumeSize is not considered.

MakeCreateVolumeReq() itself is modified to not pass LimitBytes, to match most existing cases.

Fixes some failures due to TestVolumeSize not passed to CSI.

```release-note
NONE
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -217,15 +217,7 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 			name := "sanity"
 
 			// Create a new volume.
-			req := &csi.CreateVolumeRequest{
-				Name: name,
-				VolumeCapabilities: []*csi.VolumeCapability{
-					TestVolumeCapabilityWithAccessType(sc, csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER),
-				},
-				Secrets:    sc.Secrets.CreateVolumeSecret,
-				Parameters: sc.Config.TestVolumeParameters,
-			}
-
+			req := MakeCreateVolumeReq(sc, name)
 			vol := r.MustCreateVolume(context.Background(), req)
 
 			// List volumes and check for the newly created volume.
@@ -291,15 +283,7 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 				By("creating required new volumes")
 				for i := initialTotalVols; i < minVolCount; i++ {
 					name := "sanity" + strconv.Itoa(i)
-					req := &csi.CreateVolumeRequest{
-						Name: name,
-						VolumeCapabilities: []*csi.VolumeCapability{
-							TestVolumeCapabilityWithAccessType(sc, csi.VolumeCapability_AccessMode_SINGLE_NODE_WRITER),
-						},
-						Secrets:    sc.Secrets.CreateVolumeSecret,
-						Parameters: sc.Config.TestVolumeParameters,
-					}
-
+					req := MakeCreateVolumeReq(sc, name)
 					vol, err := r.CreateVolume(context.Background(), req)
 					Expect(err).NotTo(HaveOccurred())
 					Expect(vol).NotTo(BeNil())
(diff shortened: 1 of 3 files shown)


pkg/sanity/controller.go
```diff
@@ -502,40 +438,17 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
...
-						RequiredBytes: size2,
-						LimitBytes:    size2,
-					},
-					Secrets:    sc.Secrets.CreateVolumeSecret,
-					Parameters: sc.Config.TestVolumeParameters,
-				},
-			)
+			req2 := MakeCreateVolumeReq(sc, name)
+			req2.CapacityRange = &csi.CapacityRange{
+				RequiredBytes: TestVolumeSize(sc) * 2,
```
xing-yang, line 449: I think it's better to fix MakeCreateVolumeReq to take CapacityRange as input.

huww98: We have 41 usages of MakeCreateVolumeReq now. Do we really want to make 40 invocations more complex to simplify one case?

huww98:
/remove-lifecycle stale

@xing-yang  Could you take a look at this again?

huww98:
/remove-lifecycle rotten
still relevant, waiting for review

pkg/sanity/controller.go
```diff
@@ -1640,7 +1489,6 @@ func MakeCreateVolumeReq(sc *TestContext, name string) *csi.CreateVolumeRequest
 		},
 		CapacityRange: &csi.CapacityRange{
 			RequiredBytes: size1,
-			LimitBytes:    size1,
```
jsafrane, line 1643: Will there be a test with `LimitBytes` set? This change will remove it from all current `MakeCreateVolumeReq` callers and there are quite a lot of them already.

pkg/sanity/node.go
```diff
@@ -538,21 +516,7 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
 
 			vol := r.MustCreateVolume(
 				context.Background(),
-				&csi.CreateVolumeRequest{
-					Name: name,
-					VolumeCapabilities: []*csi.VolumeCapability{
-						{
-							AccessType: &csi.VolumeCapability_Mount{
-								Mount: &csi.VolumeCapability_MountVolume{},
-							},
```
jsafrane, line 547: This is not present in `MakeCreateVolumeReq` result. Should it be there?

pkg/sanity/controller.go
huww98: I think this is not required by most CSI driver. Kubernetes does not use `LimitBytes`.

I can change `should fail when requesting to create a volume with already existing name and different capacity` test back to set LimitBytes. It needs to ensure the first volume is smaller than `size2`.

pkg/sanity/node.go
huww98: Yes, `TestVolumeCapabilityWithAccessType` is also used in `MakeCreateVolumeReq`

pkg/sanity/controller.go
huww98: Maybe we can add a dedicated test that set very small `LimitBytes` (say 1 byte), and assert the CreateVolume would fail.

jsafrane: Yes, a single test with `LimitBytes` would ensure we don't lose the test coverage. All the others can use `MakeCreateVolumeReq` as you wrote.

pkg/sanity/controller.go
```diff
@@ -492,39 +428,21 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 			By("creating a volume")
 			name := UniqueString("sanity-controller-create-twice-different")
 			size1 := TestVolumeSize(sc)
+			req1 := MakeCreateVolumeReq(sc, name)
+			req1.CapacityRange = &csi.CapacityRange{
+				RequiredBytes: size1,
+				LimitBytes:    size1, // Set Limit to ensure this volume is smaller than size2
```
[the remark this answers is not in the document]
huww98: Done. And added `Skip` when returned volume size is 0 (unknwon, for e.g. NFS share).

PR #525, opened 2024-04-02
wait for snapshot ReadyToUse before using

huww98:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

wait for snapshot ReadyToUse before using

Fixes test failure due to not ready snapshots

```release-note
NONE
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"time"
 
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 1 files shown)


pkg/sanity/controller.go
```diff
@@ -569,9 +570,20 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 				Skip("Snapshot not supported")
 			}
 
-			By("creating a snapshot")
+			By("creating a volume")
 			vol1Req := MakeCreateVolumeReq(sc, UniqueString("sanity-controller-source-vol"))
-			snap, _ := r.MustCreateSnapshotFromVolumeRequest(context.Background(), vol1Req, UniqueString("sanity-controller-snap-from-vol"))
+			vol1 := r.MustCreateVolume(context.Background(), vol1Req)
```
xing-yang, line 575: MustCreateSnapshotFromVolumeRequest is called in multiple places. Can you modify that instead of replacing it at this one place?

huww98: Only this snapshot is actually used for creating volume. So I think only this one needs fix. All other places are fine.

huww98: @xing-yang Could you PTAL again? thanks.

PR #538, opened 2024-06-03
Add ControllerModifyVolume RPC sanity tests

AndrewSirenko:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test

/kind feature
> /kind flake

Adds sanity tests for the [ControllerModifyVolume](https://github.com/container-storage-interface/spec/blob/master/spec.md#controllermodifyvolume) RPC such that VolumeAttributesClass ModifyVolume enhancement KEP-3751 can be promoted to Beta

N/A

Squashed the commits, but previously we also had:
- Bump csi-driver-host-path
- Refactor sanity with ExpectErrorCode helper
- Fix uncleaned tempfile

Active Questions: 
1. ~Should we enforce empty MutableParameters list as InvalidArgument Err? KEP says don't enforce because field OPTIONAL, csi-spec has no mention, [k/k type](https://github.com/kubernetes/kubernetes/blob/c3689b9f8b8c1a1ac2bbc4e3b3cda28d83849ec2/pkg/apis/storage/types.go#L701) says required, hostpath-driver does enforce with InvalidArgument on empty.~ Do not enforce 
2. Should we skip tests if no testvolumemutableparameters yaml file passed in?
3. ~How to enforce no conflicts between SC and VAC if both are opaque? Loop through parameter lists for duplicated keys? Or leave out test for now.~ Leave out for now, made issue #549.  

```release-note
Add csi-sanity tests for ControllerModifyVolume RPC
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ build-sanity:
 	$(MAKE) -C cmd/csi-sanity all
 
 
-TEST_HOSTPATH_VERSION=v1.7.3
+TEST_HOSTPATH_VERSION=v1.14.1
 TEST_HOSTPATH_SOURCE=bin/hostpath-source
 TEST_HOSTPATH_REPO=https://github.com/kubernetes-csi/csi-driver-host-path.git
 bin/hostpathplugin:
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -145,6 +145,7 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 				case csi.ControllerServiceCapability_RPC_PUBLISH_READONLY:
 				case csi.ControllerServiceCapability_RPC_CLONE_VOLUME:
 				case csi.ControllerServiceCapability_RPC_EXPAND_VOLUME:
+				case csi.ControllerServiceCapability_RPC_MODIFY_VOLUME:
 				case csi.ControllerServiceCapability_RPC_LIST_VOLUMES_PUBLISHED_NODES:
 				case csi.ControllerServiceCapability_RPC_GET_VOLUME:
 				case csi.ControllerServiceCapability_RPC_VOLUME_CONDITION:
@@ -199,12 +200,7 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 					StartingToken: "invalid-token",
 				},
 			)
-			Expect(err).To(HaveOccurred())
-			Expect(vols).To(BeNil())
-
-			serverError, ok := status.FromError(err)
-			Expect(ok).To(BeTrue())
-			Expect(serverError.Code()).To(Equal(codes.Aborted), "unexpected error: %s", serverError.Message())
+			ExpectErrorCode(vols, err, codes.Aborted)
 		})
 
 		It("check the presence of new volumes and absence of deleted ones in the volume list", func() {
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -38,7 +38,7 @@ build-sanity:
 	$(MAKE) -C cmd/csi-sanity all
 
 
-TEST_HOSTPATH_VERSION=v1.7.3
+TEST_HOSTPATH_VERSION=v1.13.0
```
xing-yang, line 41: There's a newer version: https://github.com/kubernetes-csi/csi-driver-host-path/releases/tag/v1.14.0

pkg/sanity/controller.go
```diff
@@ -1580,18 +1516,135 @@ var _ = DescribeSanity("ExpandVolume [Controller Server]", func(sc *TestContext)
...
+			Secrets: sc.Secrets.ControllerModifyVolumeSecret,
 		}
-		rsp, err := r.ControllerExpandVolume(context.Background(), expReq)
+		rsp, err := r.ControllerModifyVolume(context.Background(), modifyReq)
+		ExpectErrorCode(rsp, err, codes.InvalidArgument)
+	})
+
+	// TODO Q: Should we enforce Empty MutableParameters list as InvalidArgument Err?
+	// TODO Q Cont: KEP says it is OPTIONAL, k/k: storage/types says required, but no mention in CSI Spec
+	It("should fail if no mutable parameters specified", func() {
```
xing-yang, line 1579: CSI spec says mutable_parameters is optional so empty MutableParameters list is fine.
```
  // Plugin specific creation-time parameters passed in as opaque 
  // key-value pairs. These mutable_parameteres MAY also be
  // changed during the lifetime of the volume via a subsequent
  // `ControllerModifyVolume` RPC. This field is OPTIONAL.
  // The Plugin is responsible for parsing and validating these
  // parameters. COs will treat these as opaque.
  
  // Plugins MUST treat these 
  // as if they take precedence over the parameters field.
  // This field SHALL NOT be specified unless the SP has the
  // MODIFY_VOLUME plugin capability.
  map<string, string> mutable_parameters = 8 [(alpha_field) = true];
```

pkg/sanity/controller.go
```diff
@@ -1580,18 +1516,135 @@ var _ = DescribeSanity("ExpandVolume [Controller Server]", func(sc *TestContext)
...
+	var r *Resources
+
+	BeforeEach(func() {
+		r = &Resources{
+			ControllerClient: csi.NewControllerClient(sc.ControllerConn),
+			Context:          sc,
+		}
+
+		if !isControllerCapabilitySupported(r, csi.ControllerServiceCapability_RPC_MODIFY_VOLUME) {
+			Skip("ControllerExpandVolume not supported")
```
xing-yang, line 1537: typo: ExpandVolume -> ModifyVolume

pkg/sanity/controller.go
```diff
@@ -1580,18 +1516,135 @@ var _ = DescribeSanity("ExpandVolume [Controller Server]", func(sc *TestContext)
...
+		modifyReq := &csi.ControllerModifyVolumeRequest{
+			VolumeId: vol.GetVolume().GetVolumeId(),
+			MutableParameters: map[string]string{
+				"XXX_FakeKey": "XXX_FakeValue",
+			},
+			Secrets: sc.Secrets.ControllerModifyVolumeSecret,
+		}
+		rsp, err := r.ControllerModifyVolume(context.Background(), modifyReq)
+		ExpectErrorCode(rsp, err, codes.InvalidArgument)
 	})
```
xing-yang, line 1648: In the CSI spec, it says 'Values specified in mutable_parameters MUST take precedence over the values from parameters.`.  Can you add a test case for that?

AndrewSirenko: An issue I had implementing this test case pre-pr submission is that `parameters` & `mutable_parameters` are opaque, and we have no visibility into created volume respecting the precedence.

One way we could potentially test this is by finding a shared parameter + mutable_parameter (looping through the two lists), and forcing the mutable_parameter to be invalid. If the driver would return `INVALID_ARGUMENTS` error code, it suggests that precedence is respected, and the test would pass. 

However, I was afraid that this would be too 'hacky' of a test to add, because we don't know if there are shared parameters, or what an invalid parameter might be. 

---

Hemant suggested that the CSI Driver writer can pass in an additional flag where a shared parameter name can be passed, and requiring that this flag be passed if driver implements the ModifyVolume RPC. With this + the invalid_arguments precedence test above we could test this CSI Spec requirement. 

Pros:
- This forces CSI Driver writer to know about this precedence requirement. 

Cons:
- Another test flag/file. Not the most intuitive to add because it likely needs to mention what would make the parameter invalid. 
- We need some way of skipping this test if the tested CSI Driver does not share any parameters between `parameters` and `mutable_parameters`. 

---

I'll go add a commit with this test and additional flag, and then we can debate its inclusion once we see the code.

AndrewSirenko: Resolving for now in place of ^^^ issue

xing-yang:
Can you squash the commits?

PR #568, opened 2024-12-31
More tests of CreateVolume CapacityRange

ebblake:
**What type of PR is this?**
>
> /kind api-change

/kind bug

> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

While working on a CSI driver, I had several bugs in handling CapacityRange, even though my driver was passing csi-sanity, and where I think the CSI spec was clear that I should have been diagnosing improper input instead of silently accepting garbage.  While adding testsuite coverage for those cases, I also noticed a big TODO comment where the test was running two identical runs of CreateVolume against a CapacityRange of TestVolumeSize() instead of one of the tests checking for behavior with CapacityRange omitted.

Thus, this PR adds a few more test cases (rejecting negative values, rejecting the impossible RequiredSize > LimitSize, accepting a CapacityRange with LimitSize but not RequiredSize set) and modifies the existing NoCapacity test to actually omit capacity (with some flexibility on acceptable responses from the implementation).

Fixes #

```release-note
Additional tests regarding the handling of the optional CapacityRange within CreateVolume. If your driver fails the new tests, please consider improving your driver to validate more strictly, but also report the failure in case the testsuite and/or specification needs to be relaxed to support your existing behavior.
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -394,6 +394,62 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 			ExpectErrorCode(rsp, err, codes.InvalidArgument)
 		})
 
+		It("should fail with negative capacity", func() {
+
+			By("creating a volume")
+			name := UniqueString("sanity-controller-create-single-with-negative-capacity")
+
+			vol, err := r.CreateVolume(
+				context.Background(),
+				&csi.CreateVolumeRequest{
+					Name: name,
+					VolumeCapabilities: []*csi.VolumeCapability{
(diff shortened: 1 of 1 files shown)


ebblake:
Looks like csi-driver-host-path has the same two bugs as my driver did. I'll prepare a patch for that project, and link it here soon enough...

ebblake:
If https://github.com/kubernetes-csi/csi-driver-host-path/pull/577 is accepted, this PR will need to be updated to have 'make test-sanity' run against a release of the hostpath driver with those fixes.

PR #597, opened 2026-05-15
Skip controller expansion check for node-only drivers

### LINKED ISSUE #595: Node sanity setup fails on node-only drivers due to ControllerGetCapabilities call
Hello! I believe we've found a bug in the sanity test setup for drivers that implement only the Identity and Node services. 

We ran into this at Netflix with a CSI driver that only provides the Identity and Node services, which is valid under the CSI spec as long as the driver does not advertise `CONTROLLER_SERVICE` in `GetPluginCapabilities`.

## What happened?

The Node Service sanity suite's `BeforeEach` calls `ControllerGetCapabilities` unconditionally to check for `EXPAND_VOLUME`, even when the driver under test does not advertise `CONTROLLER_SERVICE` in `GetPluginCapabilities`. For a node-only driver (Identity + Node services only), this RPC fails with `Unimplemented` and the suite fails before the Node tests run.

See [`pkg/sanity/node.go#L287`](https://github.com/kubernetes-csi/csi-test/blob/b0702afc614aba2d487d090c6dcd583657f9d05a/pkg/sanity/node.go#L287):

```go
controllerExpansionSupported = isControllerCapabilitySupported(cl, csi.ControllerServiceCapability_RPC_EXPAND_VOLUME)
```

The adjacent `PUBLISH_UNPUBLISH_VOLUME` lookup a few lines above is correctly gated on `providesControllerService` ([`node.go#L279-L283`](https://github.com/kubernetes-csi/csi-test/blob/b0702afc614aba2d487d090c6dcd583657f9d05a/pkg/sanity/node.go#L279-L283)):

```go
if providesControllerService {
    controllerPublishSupported = isControllerCapabilitySupported(
        cl,
        csi.ControllerServiceCapability_RPC_PUBLISH_UNPUBLISH_VOLUME)
}
```

The `EXPAND_VOLUME` lookup right below it is missing the same guard.

## What did you expect to happen?

A driver that only implements the Identity and Node services should be able to run the Node sanity suite. The CSI spec permits this as long as the driver does not advertise `CONTROLLER_SERVICE` in `GetPluginCapabilities`.

## How to reproduce it

1. Implement a CSI driver that returns Identity and Node services only (no `CONTROLLER_SERVICE` in `GetPluginCapabilities`).
2. Run `sanity.GinkgoTest` against it.
3. `BeforeEach` fails on the unguarded `ControllerGetCapabilities` call and the Node tests never execute.

### Proposed fix

Gate the expansion lookup on `providesControllerService`, matching the pattern used for `PUBLISH_UNPUBLISH_VOLUME`:

```go
if providesControllerService {
    controllerExpansionSupported = isControllerCapabilitySupported(
        cl,
        csi.ControllerServiceCapability_RPC_EXPAND_VOLUME)
}
```

## Environment

- `csi-test` version: `v5.4.0`
- CSI spec version: 1.x (any version where Controller service is optional)
- Go version: 1.23

I would be happy to drive the fix for this!

Andre-Amor:
**What type of PR is this?**
/kind bug

The Node Service sanity suite checked `ControllerGetCapabilities` for `EXPAND_VOLUME` unconditionally during `BeforeEach`. For CSI drivers that implement only the Identity and Node services and do not advertise `CONTROLLER_SERVICE`, that RPC is expected to be unimplemented. This caused the Node sanity suite to fail before running the Node tests.

This PR gates the controller expansion capability lookup.

Fixes #595

```release-note
NONE
```

### DIFF
--- a/pkg/sanity/node.go
+++ b/pkg/sanity/node.go
@@ -284,7 +284,9 @@ var _ = DescribeSanity("Node Service", func(sc *TestContext) {
 		nodeStageSupported = isNodeCapabilitySupported(n, csi.NodeServiceCapability_RPC_STAGE_UNSTAGE_VOLUME)
 		nodeVolumeStatsSupported = isNodeCapabilitySupported(n, csi.NodeServiceCapability_RPC_GET_VOLUME_STATS)
 		nodeExpansionSupported = isNodeCapabilitySupported(n, csi.NodeServiceCapability_RPC_EXPAND_VOLUME)
-		controllerExpansionSupported = isControllerCapabilitySupported(cl, csi.ControllerServiceCapability_RPC_EXPAND_VOLUME)
+		if providesControllerService {
+			controllerExpansionSupported = isControllerCapabilitySupported(cl, csi.ControllerServiceCapability_RPC_EXPAND_VOLUME)
+		}
 		r = &Resources{
 			Context:          sc,
 			ControllerClient: cl,


carlory:
/assign @saad-ali 
for approval.

PR #618, opened 2026-08-05
Fix ListVolumes test with concurrent volume changes

### LINKED ISSUE #221: CSI sanity test check the presence of new volumes and absence of deleted ones in the volume list can fail when used with other workloads
The test (in pkg/sanity/controller.go) steps:
Step 1: The test checks how many volumes are already present:
 vols, err := c.ListVolumes(
                                context.Background(),
                                &csi.ListVolumesRequest{})
                        totalVols := len(vols.GetEntries())

Step 2: The test adds and deletes a volume

Step 3: The test checks how many volumes are present again

 vols, err = c.ListVolumes(
                                context.Background(),
                                &csi.ListVolumesRequest{})
                        Expect(len(vols.GetEntries())).To(Equal(totalVols))

The issue is that if someone else is using the server, the second time ListVolumesResponse.GetEntries is called, it is not guaranteed to be the same as when it was called the first time. The test is not accounting for other volumes being made/deleted while it runs, which can cause it to fail sometimes.

dlanov:
**What type of PR is this?**

/kind bug

Updates the `ListVolumes` sanity test to verify the presence and absence of the specific volume created by the test instead of comparing global volume counts.

This prevents unrelated concurrent volume creation or deletion from causing the test to fail. It also uses a unique volume name to avoid collisions between concurrent test runs.

Fixes #221

Local validation:

- `make test-go`
- `make test-vet`
- `make test-fmt`
- `git diff --check`

```release-note
NONE
```

### DIFF
--- a/pkg/sanity/controller.go
+++ b/pkg/sanity/controller.go
@@ -209,47 +209,55 @@ var _ = DescribeSanity("Controller Service [Controller Server]", func(sc *TestCo
 		})
 
 		It("check the presence of new volumes and absence of deleted ones in the volume list", func() {
-			// List Volumes before creating new volume.
+			containsVolume := func(response *csi.ListVolumesResponse, volumeID string) bool {
+				for _, entry := range response.GetEntries() {
+					if entry.GetVolume().GetVolumeId() == volumeID {
+						return true
+					}
+				}
+				return false
+			}
+
+			By("creating a volume")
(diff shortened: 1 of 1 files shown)


dlanov:
Superseded by #619. I created a replacement branch to remove the invalid issue reference from the commit message without force-pushing.

